PR #51, opened 2024-07-18
Remove package specifier from codegen Cargo.toml

neekolas:
## tl;dr

I was running into issues trying to use the codegen'd package:

```
error: no matching package named `core` found
```

This is because the Cargo.toml inside the `core` folder actually has the name field set to `rindexer`. Cargo cares about the name, not the folder, so it can't import.

### DIFF
--- a/core/src/generator/build.rs
+++ b/core/src/generator/build.rs
@@ -360,7 +360,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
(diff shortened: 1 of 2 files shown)


joshstevens19:
Oops I changed this last minute il review this in morning thanks for this @neekolas

joshstevens19 (approved):

joshstevens19 (approved):

joshstevens19:
Merged will release a new version shortly

joshstevens19:
deployed

PR #60, opened 2024-07-22
feat: create a minimal docker image and github workflow for building it when pushed to master branch

crebsy:
# What I did

- Created a docker image which contains `rindexer` and the graphql server binary with docker
- Github workflow that builds and tags it on ghcr.io when a branch starting with `release/` is merged/pushed to `master`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ RUN rustup target add x86_64-unknown-linux-musl
 RUN cargo build --release --target x86_64-unknown-linux-musl
 RUN strip /app/target/x86_64-unknown-linux-musl/release/rindexer_cli
 
-FROM scratch
+FROM --platform=linux/amd64 scratch
 COPY --from=builder /app/target/x86_64-unknown-linux-musl/release/rindexer_cli /app/rindexer
 
 ENTRYPOINT ["/app/rindexer"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,12 @@
 FROM --platform=linux/amd64 clux/muslrust:stable as builder
 ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
-RUN mkdir /app
 WORKDIR /app
 COPY . .
 RUN rustup target add x86_64-unknown-linux-musl
 RUN cargo build --release --target x86_64-unknown-linux-musl
 RUN strip /app/target/x86_64-unknown-linux-musl/release/rindexer_cli
 
-FROM --platform=linux/amd64 node:16 as node-builder
-RUN mkdir /app
+FROM --platform=linux/amd64 node:lts as node-builder
 WORKDIR /app
 COPY . .
 RUN cd /app/graphql && npm i && npm run build-linuxstatic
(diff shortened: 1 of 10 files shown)


joshstevens19:
good PR @crebsy - can we add some documentation?

crebsy:
@joshstevens19
I've just added a small section for Docker inside the README and also packaged the static binary for the graphql server into the final image.

Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM --platform=linux/amd64 clux/muslrust:stable as builder
+ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
+RUN mkdir /app
```
o-az, line 3:
```suggestion
```

`WORKDIR` instruction automatically creates directory if it doesn't exist

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+WORKDIR /app
+COPY . .
+RUN rustup target add x86_64-unknown-linux-musl
+RUN cargo build --release --target x86_64-unknown-linux-musl
+RUN strip /app/target/x86_64-unknown-linux-musl/release/rindexer_cli
+
+FROM --platform=linux/amd64 node:16 as node-builder
```
o-az, line 10:
```suggestion
FROM --platform=linux/amd64 node:lts as node-builder
```

Node.js v16 is no longer maintained and doesn't receive security updates
https://endoflife.date/nodejs

Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN mkdir /app
```
o-az, line 11:
```suggestion
```

o-az (changes requested):

Dockerfile
crebsy: thanks, removed!

Dockerfile
crebsy: yeah I've seen that it was used inside the package.json for building the graphql binary but you're right we should bump it

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
+WORKDIR /app
+COPY . .
+RUN rustup target add x86_64-unknown-linux-musl
+RUN cargo build --release --target x86_64-unknown-linux-musl
+RUN strip /app/target/x86_64-unknown-linux-musl/release/rindexer_cli
+
+FROM --platform=linux/amd64 node:lts as node-builder
+WORKDIR /app
+COPY . .
```
o-az, line 11: I think copying the files/dirs needed specifically will prevent unwanted files from being copied over

alternatively you can ignore everything in `.dockerignore` except for the files you want to copy over. Example

```dockerignorr
# ignore everything
*

# except the following
!scripts/
!Dockerfile
!.dockerignore
!rindexer
```

https://github.com/o-az/rindexer-railway/blob/main/.dockerignore

crebsy: true but in the end, we'll copy over only the two static binaries into the final alpine image and just publish that to the registry.

o-az:
I was able to run this locally using just the `Dockerfile`. So I think you can merge this and I will update my pr to include railway template @joshstevens19

o-az (approved):

joshstevens19 (approved):
Epic work

Issue #43, opened 2024-07-16
Alternative data sources to RPC


ozgrakkurt:
Hey,

Did you consider alternative data sources to RPC? What do you think about something like this: https://github.com/enviodev/hypersync-client-rust ?

I am working on hypersync so this is a bit of a shill but It is genuinely a massive upgrade compared to RPC on indexing cases like rindexer

joshstevens19:
Hey JSONRPC was the first release but as explained here https://rindexer.xyz/docs/introduction/what-is-rindexer#future-vision-for-rindexer we want rindexer to not care how the data is synced so for sure want to add many different data sources so would be open to seeing how we could make rindexer hypersync compatible and then editable in the yaml..?

ozgrakkurt:
guess just need to make this a trait? https://github.com/joshstevens19/rindexer/blob/a43544eaea6735acd2f64e5f94c1b4910fc8e41b/core/src/provider.rs#L101

I can work on converting it into a trait as well if it makes sense. Don't want to distrupt your flow though if you plan to change other things

joshstevens19:
I am happy to accept a PR for this for sure dude way better for you to do then me as you are building the hypersync, if its then a hyperscan node swap it out then i think we are good - go for it 💪 shout if you need any help

dzhelezov:
Would be happy to add SQD as an alternative data source too!

joshstevens19:
if its as simple as adding this trait and calling via jsonrpc as per the hyperscan URL then sounds good to me would accept a PR @dzhelezov

joshstevens19:
https://github.com/joshstevens19/rindexer/pull/103

ozgrakkurt:
> Someone has submitted a hypersync draft https://github.com/joshstevens19/rindexer/pull/103 

To clarify Misha is in hypersync team so the implementation is "official" by envio

PR #74, opened 2024-08-06
Add Envio to other-indexing-tools page

moose-code:


joshstevens19:
nice - edited the video you should get hyperscan working on rindexer if its just a case of making a trait and then making it work i would be down cc https://github.com/joshstevens19/rindexer/issues/43

joshstevens19 (approved):

moose-code:
> nice - edited the video you should get hyperscan working on rindexer if its just a case of making a trait and then making it work i would be down cc #43

Thanks! Yes we are still planning on doing this, just lots on the boil 👍

PR #76, opened 2024-08-07
docs: erpc

kasrakhosravi:
Adding [eRPC](https://github.com/erpc/erpc) as RPC proxy and caching for indexing use-cases

@joshstevens19 let me know if this makes sense to be here :)

btw, the preview feature of Vercel needs authentication

### DIFF
--- a/documentation/docs/pages/docs/start-building/yaml-config/networks.mdx
+++ b/documentation/docs/pages/docs/start-building/yaml-config/networks.mdx
@@ -48,7 +48,6 @@ project_type: no-code
 networks:
 - name: ethereum
   chain_id: 1
-  // erpc
   rpc: https://mainnet.gateway.tenderly.co // [!code focus]
 ```
 


documentation/docs/pages/docs/references/rpc-node-providers.mdx
```diff
@@ -33,4 +33,48 @@ You can use [chainlist](https://chainlist.org/) to find all the providers which
...
+
+See [a complete config example](https://docs.erpc.cloud/config/example) for inspiration.
+
+2. Use the Docker image:
+
+```bash
+docker run -v $(pwd)/erpc.yaml:/root/erpc.yaml -p 4000:4000 -p 4001:4001 ghcr.io/erpc/erpc:latest
+```
+
+3. Send your first request:
```
joshstevens19, line 66: lets remove this send your first request and talk about rindexer aka now you can now set this rpc URl in the network > rpc name https://rindexer.xyz/docs/start-building/yaml-config/networks#rpc and then you are set to go

joshstevens19, line 66: lets remove this send your first request and talk about rindexer aka now you can now set this rpc URl in the network > rpc name https://rindexer.xyz/docs/start-building/yaml-config/networks#rpc and then you are set to go

joshstevens19 (changes requested):

documentation/docs/pages/docs/references/rpc-node-providers.mdx
```diff
@@ -33,4 +33,48 @@ You can use [chainlist](https://chainlist.org/) to find all the providers which
...
 
-rindexer should work with any local nodes you run including [Anvil](https://book.getfoundry.sh/anvil/) and [Hardhat](https://hardhat.org/)
\ No newline at end of file
+rindexer should work with any local nodes you run including [Anvil](https://book.getfoundry.sh/anvil/) and [Hardhat](https://hardhat.org/)
+
+# RPC Proxy and Caching
+
+[eRPC](https://erpc.cloud/) is a fault-tolerant EVM RPC proxy and re-org aware permanent caching solution. It is built with read-heavy use-cases in mind such as data indexing and high-load frontend usage.
+
+# Quickstart
```
joshstevens19, line 42: you can also do a shoutout in https://rindexer.xyz/docs/start-building/yaml-config/networks#rpc about this as in turn this is load balancing the request as well which rindexer doesn't have by default so this can be its default

joshstevens19 (changes requested):

joshstevens19:
great PR yes it may be easier to run "npm dev" on the docs repo locally but i have allowed this to be seen now, added a few comments thanks for PR

aramalipoor:
@joshstevens19 what do you think of adding erpc in the docker-compose.yml and alsoa. basic erpc.yaml example in the root.

So basically the flow for users will be to add their RPC endpoints in erpc.yaml and then in reindexer only use erpc.

A fun idea is if you can "import" erpc.yaml from reindexer config, so users don't need to change anything there if they add a new network in erpc.yaml.

joshstevens19:
> @joshstevens19 what do you think of adding erpc in the docker-compose.yml and also a basic erpc.yaml example.
> 
> So basically the flow for users will be to add their RPC endpoints in erpc.yaml and then in reindexer only use erpc.
> 
> A fun idea is if you can "import" erpc.yaml from reindexer config, so users don't need to change anything there if they add a new network in erpc.yaml.

what if we as you said exposed a new property in the rindexer.yaml called erpc and erpc could try to also read from an rindexer.yaml as well? this means we don't need 2 yamls firstly!

I'm not fully into making the docker have it by default though as deployments etc may get a bit odd so i think we can keep that as a manual step but just get them to add a section to the docker compose?

kasrakhosravi:
@joshstevens19 let me know your thoughts about this structure

Screenshot 2024-08-08 at 10 58 54
Screenshot 2024-08-08 at 10 59 03

joshstevens19:
> @joshstevens19 let me know your thoughts about this structure
> 
> `references/rpc-node-providers` Screenshot 2024-08-08 at 10 58 54
> 
> `start-building/yaml-config/networks` Screenshot 2024-08-08 at 10 59 03

this looks good !!

joshstevens19:
let me know if this is the final version your end looks good mine - i will then merge @kasrakhosravi @aramalipoor

kasrakhosravi:
@joshstevens19 let's get this one merged in 🚀 we can later also think about an optional erpc property in `rindexer.yaml` if it makes sense

Issue #94, opened 2024-08-28
`rindexer codegen typings` leaks absolute paths


roshanr95:
As in the title, the command generates files which contain absolute paths. Are the generated files not supposed to be checked into VCS? If not, happy to make a PR that makes them relative.

joshstevens19:
Hey thanks for this yes it’s a bug feel free to fix it happily accept a PR

joshstevens19:
do you want to take this @roshanr95 if so great - i am away till next week and can fix once back if not

roshanr95:
Yes, I'll take this. I can either sanitize the absolute path before printing to file, or purge absolute paths in the codebase (i.e. `std::env::current_dir`) as much as possible if not entirely. Which option would you prefer?

joshstevens19:
yeah probably can just use `../../../` approach to be honest - awesome thanks good spot as well as would break for people when they deploy it haha

roshanr95:
`Abigen` seems to convert relative paths into absolute paths internally :cry:

roshanr95:
Ah ok, opened a couple of PRs with my existing changes btw. At a conference atm, so will take a week to get back to this in case it's still open.

joshstevens19:
hey not had time to look at these but I've spotted a few other places 
![image](https://github.com/user-attachments/assets/ba024bff-0fa7-4db6-9253-161efe81eb41)
aka in the handler itself does your fix any of that I'm not working on a project which is going to use rindexer so we can get this all fixed

joshstevens19:
should now not be valid as we bring ABI in the typings itself instead of looking at the file this just means any changes to ABI we would regen the types but that feels better anyway still one lives on in csv stuff but il fix that also

PR #98, opened 2024-09-05
Relative paths instead of absolute

roshanr95:
Replaced the absolute path obtained from `std::env::current_dir()` with the relative path `.`. 

Partially fixes an issue where the generated typings file for events contained the absolute path (#94).

Includes #97.


joshstevens19:
I dont think this fixes the main issue https://github.com/joshstevens19/rindexer/pull/102 this fixes typings leak i think 1 is still present in the csv which il fix

Issue #134, opened 2024-12-30
uint256 representation


maaft:
I have this TestEvent:
```solidity
event TestEvent(uint256 value);
``` 

which I emit like this:

```solidity
emit TestEvent(0xcbaedabc99999999000000000000000000000000000000001111111143265234);
``` 

querying this via graphql yields this:

```json
{
  "data": {
    "allTestEvents": {
      "nodes": [
        {
          "value": "-23663639920292137946688621715205637255843813578955268939319981577431001181644"
        }
      ]
    }
  }
}
```

Why is this value negative? How to get the original value?

maaft:
Some further investigation:

Contract:
```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.24;

contract Test {
    event TestEvent(string name, uint256 value);
    function test() public {
        emit TestEvent("0xcbaedabc99999999000000000000000000000000000000001111111143265234", 0xcbaedabc99999999000000000000000000000000000000001111111143265234);
        emit TestEvent("0x0000000000000000000000000000000000000000000000000000000000000001", 0x0000000000000000000000000000000000000000000000000000000000000001);
        emit TestEvent("0x0fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff", 0x0fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff);
        emit TestEvent("0xf000000000000000000000000000000000000000000000000000000000000000", 0xf000000000000000000000000000000000000000000000000000000000000000);
        emit TestEvent("0xf000000000000000000000000000000000000000000000000000000000000001", 0xf000000000000000000000000000000000000000000000000000000000000001);
    }
}

```

Indexed results:
```json
{
  "data": {
    "allTestEvents": {
      "nodes": [
        {
          "name": "0xcbaedabc99999999000000000000000000000000000000001111111143265234",
          "value": "-23663639920292137946688621715205637255843813578955268939319981577431001181644"
        },
        {
          "name": "0x0000000000000000000000000000000000000000000000000000000000000001",
          "value": "1"
        },
        {
          "name": "0x0fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff",
          "value": "7237005577332262213973186563042994240829374041602535252466099000494570602495"
        },
        {
          "name": "0xf000000000000000000000000000000000000000000000000000000000000000",
          "value": "-7237005577332262213973186563042994240829374041602535252466099000494570602496"
        },
        {
[...]

PR #135, opened 2024-12-30
fix: issue two's complement on large u256 values

joshstevens19:
fixes https://github.com/joshstevens19/rindexer/issues/134


maaft:
@joshstevens19  works for me

joshstevens19:
Cool can’t deploy atm as AFK but will when I can

PR #138, opened 2024-12-31
Update helm values

mateo-moon:
- Fix bug with graphql won't starting by increasing requested RAM to 600Mi
  - Change image to rindexer-bundled

### DIFF
--- a/helm/rindexer/values.yaml
+++ b/helm/rindexer/values.yaml
@@ -1,6 +1,6 @@
 replicaCount: 1
 image:
-  repository: ghcr.io/joshstevens19/rindexer
+  repository: ghcr.io/joshstevens19/rindexer-bundled
   tag: "latest"
   pullPolicy: IfNotPresent
 
@@ -18,11 +18,13 @@ ingress:
 
 resources:
   limits:
-    cpu: 100m
-    memory: 128Mi
+    cpu: 200m
+    memory: 800Mi 
   requests:
     cpu: 100m
-    memory: 128Mi
+    # Minimum amount for graphql. Otherwise it won't start
+    # The actual memory usage is a lot less than this
+    memory: 600Mi 
 
 postgresql:
   enabled: false  # Enable or disable PostgreSQL integration


joshstevens19:
Thanks for PR @mateo-moon - Can you cross check the docs: 

https://github.com/joshstevens19/rindexer/blob/master/documentation/docs/pages/docs/deploying/aws.mdx#22-configure-the-valuesyaml-file

https://github.com/joshstevens19/rindexer/blob/master/documentation/docs/pages/docs/deploying/gcp.mdx#22-configure-the-valuesyaml-file

And fix anything related like the image and any mention of memory etc (not sure it does)

mateo-moon:
> Thanks for PR @mateo-moon - Can you cross check the docs:
> 
> https://github.com/joshstevens19/rindexer/blob/master/documentation/docs/pages/docs/deploying/aws.mdx#22-configure-the-valuesyaml-file
> 
> https://github.com/joshstevens19/rindexer/blob/master/documentation/docs/pages/docs/deploying/gcp.mdx#22-configure-the-valuesyaml-file
> 
> And fix anything related like the image and any mention of memory etc (not sure it does)

Sure. Also we use different rindexer.yaml obviously. Let me know if you want me to add changes for helm chart to be generic about rindexer.yaml.

joshstevens19:
> > Thanks for PR @mateo-moon - Can you cross check the docs:
> > https://github.com/joshstevens19/rindexer/blob/master/documentation/docs/pages/docs/deploying/aws.mdx#22-configure-the-valuesyaml-file
> > https://github.com/joshstevens19/rindexer/blob/master/documentation/docs/pages/docs/deploying/gcp.mdx#22-configure-the-valuesyaml-file
> > And fix anything related like the image and any mention of memory etc (not sure it does)
> 
> Sure. Also we use different rindexer.yaml obviously. Let me know if you want me to add changes for helm chart to be generic about rindexer.yaml.

Yeah our yaml is just to get people setup we expect people to change it but if you stop something in the helm which can be more generic for sure!

joshstevens19:
@mateo-moon how you getting on with this?

mateo-moon:
> @mateo-moon how you getting on with this?

Hey, @joshstevens19, I ran more tests and found that the numbers I came up with were not correct. There is no consistency in how much is good to set. One run might work with 500Mi, but the next could fail. The only consistent behavior I’ve observed is when no resources are defined.

I don’t know why this is happening. I’m using AWS instances in an EKS cluster, so maybe it’s related to that. But I don’t have time right now to test it properly.

I’m unsure how to approach this. Omitting resources isn’t a good practice—it feels like a hack. From my perspective, the right approach would be to identify the cause of the issue and fix it. As a quick fix, it’s possible to remove the resources or leave a comment in the values and documentation. But again, that’s just a hack.

What’s your take?

joshstevens19:
> > @mateo-moon how you getting on with this?
> 
> Hey, @joshstevens19, I ran more tests and found that the numbers I came up with were not correct. There is no consistency in how much is good to set. One run might work with 500Mi, but the next could fail. The only consistent behavior I’ve observed is when no resources are defined.
> 
> I don’t know why this is happening. I’m using AWS instances in an EKS cluster, so maybe it’s related to that. But I don’t have time right now to test it properly.
> 
> I’m unsure how to approach this. Omitting resources isn’t a good practice—it feels like a hack. From my perspective, the right approach would be to identify the cause of the issue and fix it. As a quick fix, it’s possible to remove the resources or leave a comment in the values and documentation. But again, that’s just a hack.
> 
> What’s your take?

Sorry just seeing this now - i think the issue here is that depending on what your indexing and how much data is coming in the memory may go up and down a bit same with the graphQL API i am really not sure best route here - closing this for now we can reopen if we think we can really recommend anything more to people then just "adjust based on your usage"

Issue #146, opened 2025-01-02
Append "topic_hash" to events


0xSokka:
Would be useful to append either a "topic_hash" or "topic_id" to both evens streamed and stored to easily identify/fetch them.

PR #147, opened 2025-01-02
chore: Forward event topic to events

0xSokka:
Solves #146

### DIFF
--- a/core/src/event/message.rs
+++ b/core/src/event/message.rs
@@ -6,6 +6,6 @@ use serde_json::Value;
 pub struct EventMessage {
     pub event_name: String,
     pub event_data: Value,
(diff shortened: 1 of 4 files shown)


core/src/event/message.rs
```diff
@@ -5,5 +5,6 @@ use serde_json::Value;
 pub struct EventMessage {
     pub event_name: String,
     pub event_data: Value,
+    pub event_topic: String,
```
joshstevens19, line 8: i was wondering if event_topic could stay as Signature i think it has a serde on it and should then work on create_chunk_message_raw and create_chunk_message_json (these are both the places it constructs ready to send)

joshstevens19 (approved):

PR #163, opened 2025-03-04
Alias StreamEvent Event Name

jaggad:
Adds support for aliasing an `Event` with an alternative name.

Useful for when we have SNS topics which fanout to many queues, including "shared queues" which have subscriptions/bindings to many overlapping fanout notficiations of the same name.

We want to be able to have an enum of possible events as per below and that isn't possbile without aliasing the discriminator key (`event_name` in this case).

```rs
#[derive(Deserialize, Clone, Debug, Serialize)]
#[serde(tag = "event_name")]
pub enum Events {
    Transfer { network: Network, event_data: Vec<Transfer> },
    // Note the event name `Erc721Transfer` which is mapped from the ERC721 ABI's `Transfer` event.
    // It's more pleasant to work with discriminated enums where possible.
    Erc721Transfer { network: Network, event_data: Vec<Erc721Transfer> },
}
```

**Additionally**

I'm happy to roll this out to other stream sources as well if we intend to. This PR was largely to suit my needs, and also to validate this is something we want in Rindexer


documentation/docs/pages/docs/start-building/streams/sns.mdx
```diff
@@ -374,6 +374,44 @@ contracts:
...
+        session_token: ${AWS_SESSION_TOKEN}
+        # endpoint_url is optional
+        endpoint_url: ${ENDPOINT_URL}
+      topics: // [!code focus]
+        - topic_arn: "arn:aws:sns:us-east-1:664643779377:test"
+          networks:
+            - ethereum
+          events: // [!code focus]
+            - event_name: Transfer // [!code focus]
+              alias: RocketPoolTransfer
```
joshstevens19, line 412: need a // [!code focus] here

core/src/streams/clients.rs
```diff
@@ -253,11 +253,23 @@ impl StreamsClients {
...
                 );
 
+                let alias_name = config
+                    .events
+                    .iter()
+                    .find(|n| n.event_name == event_message.event_name)
+                    .and_then(|n| n.alias.clone());
+
+                let event_message = EventMessage {
+                    event_name: alias_name.unwrap_or(event_message.event_name.clone()),
```
joshstevens19, line 263: `alias_name.unwrap_or_else` tend to be standard clippy

core/src/streams/clients.rs
```diff
@@ -253,11 +253,23 @@ impl StreamsClients {
                     chunk,
                 );
 
+                let alias_name = config
+                    .events
+                    .iter()
+                    .find(|n| n.event_name == event_message.event_name)
+                    .and_then(|n| n.alias.clone());
+
+                let event_message = EventMessage {
```
joshstevens19, line 262: wonder if we call this publish_message as the ..event_message.clone() before the let is a bit confusing aka can look like your setting itself to itself

core/src/streams/clients.rs
```diff
@@ -253,11 +253,23 @@ impl StreamsClients {
                     chunk,
                 );
 
+                let alias_name = config
```
joshstevens19, line 256: may as well support this on all the streams for sure if its just this same change

core/src/streams/clients.rs
jaggad: `publish_message` is used a few lines down for another value, I like to shadow rust varaiables sometimes to indicate the shadowed variable is no longer needed or relevant but I'm happy to call this another name say `aliased_message`.

core/src/streams/clients.rs
jaggad: Yes, `unwrap_or_else` is what I want here 👍

core/src/streams/clients.rs
jaggad: Completed, was easy to support.

core/src/streams/clients.rs
jaggad: I've made the approach generic so this is no longer relevant

core/src/streams/clients.rs
```diff
@@ -173,9 +173,24 @@ impl StreamsClients {
...
 
-    fn create_chunk_message_raw(&self, event_message: &EventMessage, chunk: &[Value]) -> String {
+    /// Gets event name, which may be an optional alias, or the contract's event name.
+    fn get_event_name(&self, events: &Vec<StreamEvent>, event_message: &EventMessage) -> String {
+        let alias_name = events
+            .iter()
+            .find(|n| n.event_name == event_message.event_name)
+            .and_then(|n| n.alias.clone());
+
+        alias_name.unwrap_or(event_message.event_name.clone())
```
jaggad, line 183:
```suggestion
        alias_name.unwrap_or_else(|| event_message.event_name.clone())
```

joshstevens19 (approved):

Issue #31, opened 2024-07-09
Support Reth nodes to index using ExExes


joshstevens19:
Reth now support execution extensions we should support ability to config an Reth node and stream data using the ExExes instead of over JSONRPC

https://www.paradigm.xyz/2024/05/reth-exex

zerosnacks:
For additional context, a major benefit of ExEx is re-org awareness meaning you can live index rather than maintaining a probabilistic safe distance

There is also now native support for backfilling, see https://github.com/paradigmxyz/reth-exex-examples/tree/main/backfill

SkandaBhat:
interested to work on this! @joshstevens19 this should be straightforward, do you mind if I take a crack at this?

joshstevens19:
Hey @SkandaBhat would love someone to take this so yeah please do it 🙏 shouldn’t be to complex main thing here is: 

- as this has to happen on startup of reth we should design it so the CLI makes it easy to run the commands as it seems you basically have to run reth under hood so for example rindexer start should also start reth up to set the ExEx connection
- any config for reth should live in the rindexer yaml making it feel like one 
- we should reuse everything rindexer does handle wise but just skipping the fetching and using reth to fetch but the rest should hit the main logic 
- we should make sure we backport the data as well and not just index the latest stuff 

shout if you need any help look forward to the PR

joshstevens19:
I also going to test and merge this in the morning just FYI https://github.com/joshstevens19/rindexer/pull/103

SkandaBhat:
Looking into this today, should have a draft early next week @joshstevens19

joshstevens19:
look forward to the PR @SkandaBhat

SkandaBhat:
@joshstevens19 sorry about this - ended up getting caught up with something else and haven't been able to wrap this up. I might be able to look into this early next week, but if someone else is willing to take this up, please go ahead and assign them.

joshstevens19:
sure @SkandaBhat think the PR would be super great so yeah feel free to pick it up next week

SkandaBhat:
@joshstevens19 Quick question here - do we want to run reth in-process? 

We could probably keep rindexer lightweight and instead use exex remotely - https://github.com/paradigmxyz/reth-exex-examples/blob/1f74410740ac996276a84ee72003f4f9cf041491/remote/bin/exex.rs

PR #166, opened 2025-03-06
Reth ExEx Support

SkandaBhat:
WIP draft for #31 

so far we have
- exex that receives backfill requests and also streams live blocks
- instead of using RPC in `fetch_logs_stream`, we instead send backfill and live stream requests to the exex, and filter logs based on addresses/topics etc and return to the same tx that's sent in the `fetch_logs_stream` fn
- helper functions to convert alloy types to ethers and vice versa (`to_alloy_filter`,  `WrappedLog::from_alloy_log`)


core/src/indexer/fetch_logs.rs
```diff
@@ -36,99 +39,271 @@ pub fn fetch_logs_stream(
...
+                    false,
+                ));
+            }
+        }
+    }
+
+    None
+}
+
+async fn process_exex_stream(
```
SkandaBhat, line 188: lots of code repetitions in this fn, needs to be improved

core/src/manifest/network.rs
```diff
@@ -24,4 +26,20 @@ pub struct Network {
...
+/// Configuration for Reth node and ExEx
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RethConfig {
+    /// Whether to enable Reth integration
+    #[serde(default)]
+    pub enabled: bool,
+
+    /// Path to the Reth data directory
+    #[serde(default)]
+    pub data_dir: Option<PathBuf>,
```
SkandaBhat, line 44: need additional config here, ref https://reth.rs/run/config.html

core/src/event/rindexer_event_filter.rs
```diff
@@ -130,4 +135,58 @@ impl RindexerEventFilter {
     pub fn raw_filter(&self) -> &Filter {
         &self.filter
     }
+
+    pub fn to_alloy_filter(&self) -> AlloyFilter {
```
SkandaBhat, line 139: add couple of tests here to verify conversion is proper

core/src/indexer/fetch_logs.rs
```diff
@@ -36,99 +39,271 @@ pub fn fetch_logs_stream(
...
-        if max_block_range_limitation.is_some() {
-            current_filter = current_filter.set_to_block(calculate_process_historic_log_to_block(
-                &from_block,
-                &snapshot_to_block,
-                &max_block_range_limitation,
-            ));
-            warn!(
+        // if we have a, exex backfill tx, we need to process the exex stream instead of using the
+        // rpc provider
+        if let Some(backfill_tx) = backfill_tx {
```
SkandaBhat, line 44: maybe there's a better way to check if exex mode is enabled. checking if backfill_tx exists is quite implicit

core/src/provider.rs
```diff
@@ -34,6 +36,58 @@ pub struct WrappedLog {
     pub block_timestamp: Option<U256>,
 }
 
+impl WrappedLog {
+    /// Creates a `WrappedLog` from an `alloy::Log` with an optional block timestamp.
+    pub fn from_alloy_log(
```
SkandaBhat, line 41: needs tests to confirm it works as expected

cli/src/commands/new.rs
```diff
@@ -141,6 +149,19 @@ pub fn handle_new_command(
             compute_units_per_second: None,
             max_block_range: None,
             disable_logs_bloom_checks: None,
+            reth: Some(RethConfig {
+                enabled: true,
+                cli_config: RethCliConfig {
```
joshstevens19, line 154: most of these could be asked i guess

core/src/indexer/fetch_logs.rs
```diff
@@ -36,99 +39,249 @@ pub fn fetch_logs_stream(
...
+                )
+                .await;
+            }
         }
     });
 
     UnboundedReceiverStream::new(rx)
 }
 
+async fn process_exex_stream(
```
joshstevens19, line 147: maybe worth spinning out an reth.rs file to hold these things in may be easier to find it etc?

core/src/indexer/fetch_logs.rs
```diff
@@ -36,99 +39,249 @@ pub fn fetch_logs_stream(
...
+
+    let (response_tx, response_rx) = oneshot::channel();
+
+    let res = reth_tx.send(ExExRequest::Start {
+        mode: ExExMode::HistoricOnly,
+        filter: filter.clone(),
+        response_tx,
+    });
+    if let Err(e) = res {
+        error!(
```
joshstevens19, line 169: what happens in this case?

core/src/indexer/fetch_logs.rs
```diff
@@ -36,99 +39,249 @@ pub fn fetch_logs_stream(
...
+            "{} - {} - Failed to start backfill",
+            config.info_log_name,
+            IndexingEventProgressStatus::Syncing.log(),
+        );
+        return;
+    };
+    info!("Backfill started for job {}", job_id);
+    let mut batched_stream = UnboundedReceiverStream::new(rx).chunks(100);
+    while let Some(logs) = batched_stream.next().await {
+        let mut from_block: u64 = u64::MAX;
```
joshstevens19, line 191: is there ever a place where you understand that you got all the backfill to say block x aka like how we create the internal rindexer db tables to check last synced block number to avoid going through them all on every startup?

core/src/manifest/network.rs
```diff
@@ -24,4 +27,101 @@ pub struct Network {
...
+    /// CLI configuration for the Reth node
+    #[serde(default)]
+    pub cli_config: RethCliConfig,
+}
+
+/// CLI configuration for the Reth node
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct RethCliConfig {
+    /// Path to the Reth data directory
+    pub data_dir: Option<PathBuf>,
```
joshstevens19, line 52: is this actual optional?

core/src/manifest/network.rs
```diff
@@ -24,4 +27,101 @@ pub struct Network {
...
+    pub enabled: bool,
+
+    /// CLI configuration for the Reth node
+    #[serde(default)]
+    pub cli_config: RethCliConfig,
+}
+
+/// CLI configuration for the Reth node
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+pub struct RethCliConfig {
```
joshstevens19, line 50: are all these fields optional in the reth solution - IMHO being strong for them to define it over putting fake values in the yaml would be more sense

core/src/reth/exex.rs
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    /// Handles the given notification and calls [`process_committed_chain`] for a committed
+    /// chain, if any.
+    async fn handle_notification(&self, notification: ExExNotification) -> eyre::Result<()> {
+        match &notification {
+            ExExNotification::ChainCommitted { new } => {
+                info!(committed_chain = ?new.range(), "Received commit");
+            }
+            ExExNotification::ChainReorged { old, new } => {
```
joshstevens19, line 93: we should do reorgs later in the whole rindexer project

core/src/start.rs
```diff
@@ -122,6 +125,19 @@ pub async fn start_rindexer(details: StartDetails<'_>) -> Result<(), StartRindex
...
 
+            // Channels for Reth block data
+            let mut reth_channels = RethChannels::new();
+
+            // Start Reth nodes for enabled networks
+            for network in manifest.reth_enabled_networks() {
+                let reth_cli = network.reth.as_ref().unwrap().cli_config.to_reth_cli();
+                info!("Starting Reth node for network: {}", network.name);
+                let reth_tx = start_reth_node_with_exex(reth_cli)?;
+                info!("Started Reth node for network: {}", network.name);
```
joshstevens19, line 136: will reth native node logs come at same time as rindexer? could you see that being a pain or easy to see both mixed?

documentation/docs/pages/docs/advanced/using-reth-exex.md
```diff
@@ -0,0 +1,35 @@
...
+      authrpc_jwtsecret: /path/to/jwt.hex // [!code focus]
+      authrpc_addr: 127.0.0.1 // [!code focus]
+      authrpc_port: 8551 // [!code focus]
+      full: false // [!code focus]
+      metrics: null // [!code focus]
+      chain: mainnet // [!code focus]
+      http: true // [!code focus]
+```
+
+## Getting Started
```
joshstevens19, line 31: would have proper instructions here now once we bring in the --reth create command

documentation/vocs.config.tsx
```diff
@@ -190,7 +190,7 @@ export default defineConfig({
       items: [
         { text: "From A Foundry Project - coming soon" },
         { text: "From A Hardhat Project - coming soon" },
-        { text: "Using Reth ExExes - coming soon" },
+        { text: "Using Reth ExExes", link: "/docs/advanced/using-reth-exex" },
```
joshstevens19, line 193: so i would say now this is a good standalone guide but the standalone guide should make it super easy for people to get up and running alongside we should have callouts in the https://rindexer.xyz/docs/start-building/create-new-project that --reth flag exists etc dont want this hidden away

cli/src/commands/new.rs
```diff
@@ -96,6 +96,14 @@ pub fn handle_new_command(
         postgres_docker_enable = postgres_docker == "yes";
     }
 
+    let mut reth_enabled = false;
+    let reth_choice = prompt_for_input_list(
```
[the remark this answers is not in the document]
SkandaBhat: makes sense!

core/src/indexer/fetch_logs.rs
SkandaBhat: last synced block still is being written because we are still returning wrapped logs back to the stream. So the `fetch_logs_stream` would restart with the last synced block from the config, as per existing functionality. So here the stream would restart with the last synced block itself

core/src/indexer/last_synced.rs
```diff
@@ -251,18 +251,16 @@ pub fn update_progress_and_last_synced_task(
...
-            .await
+                });
+            if let Err(e) =
+                update_last_synced_block_number_for_file(&config, &full_path, to_block).await
             {
                 error!(
-                    "Error updating last synced block to CSV - path - {} error - {:?}",
-                    csv_details.path, e
+                    "Error updating last synced block to CSV - path - {} error - {:?} full path - {}",
+                    csv_details.path, e, full_path.display()
```
SkandaBhat, line 263: I can maybe revert this changes but seems worthy for debugging purpose @joshstevens19 please resolve this comment if you think we can keep it

core/src/reth/exex.rs
SkandaBhat: Yes! Kept some empty logs here to make sure it's extendable. But I'll add a TODO comment here as well.

core/src/start.rs
SkandaBhat: It didnt hurt dev experience much during my testing, since it is pretty useful to track exex logs inside of reth as well. But the user could configure reth to write logs to a file instead (using cli args) and avoid seeing them.

documentation/vocs.config.tsx
SkandaBhat: Makes sense!

rindexer_rust_playground/src/rindexer_lib/typings/rindexer_playground/events/rocket_pool_eth.rs
```diff
@@ -145,9 +145,9 @@ where
         Fut: Future<Output = EventCallbackResult<()>> + Send + 'static,
     {
         let csv = AsyncCsvAppender::new(
-            r"/Users/jackedgson/Development/avara/rindexer/rindexer_rust_playground/generated_csv/RocketPoolETH/rocketpooleth-approval.csv",
+            r"/Users/skanda/Projects/rindexer/cli/../rindexer_rust_playground/generated_csv/RocketPoolETH/rocketpooleth-approval.csv",
```
SkandaBhat: @joshstevens19 not sure about this one but I ran the makefile for codegen typings because my changes needed to regenerate it (I added state_notifications into NetworkContract)

both prev and current strings are dev directory lists so maybe this can go in as it is?

core/src/indexer/process.rs
```diff
@@ -173,6 +172,7 @@ async fn process_contract_events_with_dependencies(
                     ))?;
 
                 // forces live indexing off as it has to handle it a bit differently
+                // Note: Dependencies don't support notifications yet
```
joshstevens19: dependency indexing is a common thing to do we saying this integration wont handle it?

SkandaBhat: Not sure why I left this, I'll check. Probably a remnant of older code

SkandaBhat: Confirmed that this is from older code, this comment can be removed.

PR #168, opened 2025-03-14
Debug Trace Block - Native Balance Indexing

jaggad:
I've got the native transfer indexing to a point I'm pretty happy with it for review:
Includes
- Manifest parsing for simple and complex variants, including tests
- Full stream/sink support - Postgres, Chatbot, Csv, Streams
- Reorg safe distance block trailing when enabled
- An example project in the rindexer examples fodler

Once current state is approved I will add Documentation (since I want code to be stable first to prevent wasted time on docs)

I have tested it relatively extensively and things are working smooth and well, I would like to test it a little more before we merge it but it's definitely ready for review in it's current state at least so I can get any further feedback to action.


documentation/RFC.md
```diff
@@ -0,0 +1,79 @@
...
+    network: ethereum
+    start_block: 18600000
+    end_block: 18718056  # Optional, for historic indexing
+    trace_config:
+      tracer: "call_tracer" # Or other available tracers: "pre_state_tracer", defaults to "call_tracer"
+      only_top_call: true # [default: true] When set to true, this will only trace the primary (top-level) call and not any sub-calls.
+    conditions:
+      - "from": "0x0338ce5020c447f7e668dc2ef778025ce3982662 || 0x0338ce5020c447f7e668dc2ef778025ce398266d"
+      - "value": ">0"
+      - "gas": "0"
```
jaggad, line 71: Too granular for conditions atm

- Address filter
- Value filtering

cli/output-2.json
```diff
@@ -0,0 +1,1316 @@
+{
```
0xSokka, line 1: is this file meant to be pushed?

core/src/indexer/start.rs
```diff
@@ -101,6 +124,177 @@ pub async fn start_indexing(
 
     let mut processed_network_contracts: Vec<ProcessedNetworkContract> = Vec::new();
 
+    if manifest.has_enabled_native_transfers() {
+        // We could do this inside the `no_code` setup as well
+        let providers = CreateNetworkProvider::create(manifest)
+            .expect("handle this createnetwork ERR")
+            .into_iter()
+            .map(|p| (p.network_name, p.client))
```
0xSokka, line 132: can we avoid tuple use here?

core/src/indexer/start.rs
```diff
@@ -101,6 +124,177 @@ pub async fn start_indexing(
...
+                while let Some(block_number) = block_rx.recv().await {
+                    info!("Recv block: {}. {} remaining.", block_number, block_rx.len());
+
+                    let trace_call = provider.debug_trace_block_by_number(block_number).await?;
+
+                    let native_transfers = trace_call
+                        .into_iter()
+                        .enumerate()
+                        .filter_map(|(idx, transfer)| {
+                            let is_call = transfer.result.typ == "CALL";
```
0xSokka, line 218: `typ`?

core/src/manifest/core.rs
```diff
@@ -80,20 +89,44 @@ pub struct Manifest {
...
+    /// Check if the manifest has opted-in to indexing native transfers. It is off by default.
+    pub fn has_enabled_native_transfers(&self) -> bool {
+        self.native_transfers.enabled
+    }
+
+    /// Determine if native transfers are opted in to live indexing as well. This would happen any
+    /// time an end-block is not specified for any network.
+    pub fn has_native_transfers_live_indexing(&self) -> bool {
+        let enabled = self.native_transfers.enabled;
+        let any_network_is_live = self
```
0xSokka, line 113: use `is_empty` rather than comparing usize?

core/src/streams/clients.rs
```diff
@@ -435,7 +443,17 @@ impl StreamsClients {
...
+                        config.events.iter().any(|e| e.event_name == event_message.event_name);
+
+                    // TODO
+                    //
+                    // We need to determine how we want to handle "Native_Transfer" Events.
+                    // We could allow the user to alias it by giving the event a name and just pick
+                    // it. Then we would just want to ensure it's only "one"
+                    // event name defined, or pick the first. Otherwise default
+                    // to "NativeTransfer" as the event name.
+                    if (is_user_event || is_trace_event) &&
```
0xSokka, line 456: Agreed. Letting the user pick is the the best choice, `NativeTransfer` fallback makes sense.

core/src/provider.rs
```diff
@@ -34,6 +38,26 @@ pub struct WrappedLog {
...
+pub struct TraceCall {
+    pub from: Address,
+    pub gas: String,
+    #[serde(rename = "gasUsed")]
+    pub gas_used: U256,
+    pub to: Address,
+    pub input: Bytes,
+    pub value: U256,
+    #[serde(rename = "type")]
+    pub typ: String,
```
0xSokka, line 51: Gotta love reserve words. Let's rename something like `call_type` rather than `typ`.


```suggestion
    pub call_type: String,
```

cli/output-2.json
```diff
@@ -0,0 +1,1316 @@
...
+  "jsonrpc": "2.0",
+  "id": 1,
+  "result": [
```
joshstevens19, line 4: lets gitignore this

cli/src/commands/new.rs
```diff
@@ -164,6 +165,8 @@ pub fn handle_new_command(
             streams: None,
             chat: None,
         }],
+        // TODO
+        native_transfers: NativeTransfers::default(),
```
joshstevens19, line 169: default should be off i think

core/src/indexer/native_transfer.rs
```diff
@@ -0,0 +1,130 @@
...
+        .into_iter()
+        .flatten()
+        .filter_map(|trace| {
+            let action = match trace.action {
+                Action::Call(call) => Some(call),
+                _ => None,
+            }?;
+
+            let has_value = !action.value.is_zero();
+            let no_input = action.input == Bytes::from_str("0x").unwrap();
```
joshstevens19, line 48: cant you do 

```
let no_input = action.input == Bytes::new(); 
```

joshstevens19: not sure what input is but probs nicer way to do it then bringing in an unwrap

core/src/indexer/native_transfer.rs
```diff
@@ -0,0 +1,130 @@
...
+                    return None;
+                }
+
+                Some(NativeTransfer {
+                    from: action.from,
+                    to: action.to,
+                    value: action.value,
+                    transaction_information: TxInformation {
+                        network: network_name.to_owned(),
+                        address: Address::zero(),
```
joshstevens19, line 63: you dont know the contract address this did the action from if present? could be useful say "get me all the eth transfers which were sent from x contract" or something

core/src/indexer/native_transfer.rs
```diff
@@ -0,0 +1,130 @@
...
+                        block_number: U64::from(trace.block_number),
+                        block_timestamp: None,
+                        transaction_hash: trace.transaction_hash.expect("checked prior"),
```
joshstevens19, line 66: 100% cant ever throw? if so maybe be worth just handling some some info if it does

core/src/indexer/native_transfer.rs
```diff
@@ -0,0 +1,130 @@
...
+        let from_block = native_transfers
+            .first()
+            .map(|b| b.transaction_information.block_number)
+            .unwrap_or_default();
+        let to_block = native_transfers
+            .last()
+            .map(|b| b.transaction_information.block_number)
+            .unwrap_or_default();
+
+        let stream_id = format!(
```
joshstevens19, line 94: may be worth now having a type for this as no-code uses this also and may as well share how its constructed

jaggad:
Working on refactor to top CSV, Postgres, etc here: https://github.com/jaggad/rindexer/pull/1

It's rather difficult since a lot is built around the Contract -> Event setup, but i'll try and make it work.

This would give us all of the Stream, Csv, Postgres handling with minimal extra boilerplate.

cli/output-2.json
jaggad: Nope.

core/src/indexer/start.rs
jaggad: Why? A tuple is needed to reduce it into a hashmap

core/src/manifest/core.rs
jaggad: Needs to be specific filter first, I guess we could use is_empty yeah

core/src/provider.rs
jaggad: I mimicked what ethers-rs already used which is `typ`, but this has been refactored now.

cli/output-2.json
jaggad: Yeah that's mb, will just delete it.

cli/src/commands/new.rs
jaggad: Default is off, it will internally flip to `enabled: false` at the moment.

core/src/indexer/native_transfer.rs
jaggad: I wanted to `expect` here so it would through if i encounter anything, I am under the impression it's never None so wanted to see.

core/src/indexer/native_transfer.rs
jaggad: Hmm, probably can yeah. Let me try

core/src/indexer/native_transfer.rs
jaggad: Hmm, so would you suggest making this `address` equal to the `transfer.from` address in this case?

jaggad:
So I think the approach is a fine enough one, I merged the WIP state.

Basically we just treat `indexer_evm_traces` as the contract (open to alternative names) and event name as `native_token_transfer`

So it would produce pg tables like below..

```sql
table indexer_evm_traces.native_token_transfer limit 5;
```

Overall I have a much better grasp of how to approach this problem now.

I still think we ultimately want to move in a direction where a user can tap into the raw traces themselves but this is a good start. We have all the data sinks in place and I should be able to take the learnings and do this properly now.

STRUCTURE.md
```diff
@@ -0,0 +1,107 @@
+# Structure
+
+## Cli flow
```
jaggad, line 3: Parts of this are incorrect, I can delete it unless @joshstevens19 you'd like to keep this around for future contributors.

core/src/event/callback_registry.rs
```diff
@@ -210,3 +226,130 @@ impl EventCallbackRegistry {
...
+        format!("{}::{}", self.indexer_name, self.event_name)
+    }
+}
+
+#[derive(Clone, Default)]
+pub struct TraceCallbackRegistry {
+    pub events: Vec<TraceCallbackRegistryInformation>,
+}
+
+impl TraceCallbackRegistry {
```
joshstevens19, line 294: seems a direct copy and paste of EventCallbackRegistry maybe we can share this a bit so we dont have 2 places to manage the code?

core/src/event/contract_setup.rs
```diff
@@ -90,6 +94,66 @@ impl ContractInformation {
...
+    pub fn create(
+        native_transfers: NativeTransfers,
+        network_providers: &[CreateNetworkProvider],
+    ) -> Result<TraceInformation, CreateContractInformationError> {
+        let mut details = vec![];
+        let trace_networks = native_transfers.networks.unwrap_or_default();
+
+        for n in &trace_networks {
+            let name = n.network.clone();
+            let provider = network_providers.iter().find(|item| item.network_name == name);
```
joshstevens19, line 129: now we have `network_providers.iter().find(|item| item.network_name == name)` twice lets do a nice helper again sharing the code as much as we can

core/src/indexer/start.rs
```diff
@@ -68,39 +70,214 @@ pub enum StartIndexingError {
...
     let mut dependency_event_processing_configs: Vec<ContractEventsDependenciesConfig> = Vec::new();
+
     // if you are doing advanced dependency events where other contracts depend on the processing of
     // this contract you will need to apply the dependency after the processing of the other
     // contract to avoid ordering issues
     let mut apply_cross_contract_dependency_events_config_after_processing = Vec::new();
-
     let mut processed_network_contracts: Vec<ProcessedNetworkContract> = Vec::new();
 
+    for event in trace_registry.events.iter() {
```
joshstevens19, line 187: hmm not ideal this is blocking the registry.events we should probs make them run at same time so its not blocking anything

core/src/manifest/native_transfer.rs
```diff
@@ -0,0 +1,159 @@
...
+
+fn default_enabled() -> bool {
+    true
+}
+
+/// The normalized 'Native Transfers' config.
+#[derive(Debug, Default, Serialize, Deserialize, Clone)]
+pub struct NativeTransfers {
+    #[serde(default = "default_enabled")]
+    pub enabled: bool,
```
joshstevens19, line 62: dont think you need this anymore due to the enum type?

core/src/manifest/native_transfer.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    /// None has a special meaning of "All" networks for this case. This is because we want the
+    /// functionality to be as simple to opt in to as possible.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub networks: Option<Vec<NativeTransferDetails>>,
+
+    /// For now `NativeTokenTransfer` must be the defined "Event" name.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub streams: Option<StreamsConfig>,
```
joshstevens19, line 71: make sure streams docs are also added about native

core/src/manifest/native_transfer.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub chat: Option<ChatConfig>,
```
joshstevens19, line 74: make sure streams docs are also added about native

examples/rindexer_native_transfers/rindexer.yaml
```diff
@@ -0,0 +1,40 @@
...
+          networks:
+            - ethereum
+
+          # IMPORTANT:
+          #
+          # For now we must enforce the event name is `NativeTokenTransfer`. This is
+          # because areas of the code check that the event is defined in the config on publish, but we are
+          # enforcing that the name must be: "NativeTokenTransfer".
+          events:
+            - event_name: NativeTokenTransfer
```
joshstevens19, line 31: we should abstract this away i think - this is an implementation detail we should make this bomb proof

joshstevens19 (changes requested):
great PR some comments

core/src/database/postgres/generate.rs
```diff
@@ -192,6 +195,38 @@ pub fn generate_tables_for_indexer_sql(
...
+            )?;
+
+            sql.push_str(&generate_event_table_sql_with_comments(
+                &event_names,
+                &contract_name,
+                &schema_name,
+                event_matching_name_on_other,
+            ));
+        }
+        // we still need to create the internal tables for the contract
```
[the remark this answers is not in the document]
jaggad: Not required

core/src/database/postgres/generate.rs
```diff
@@ -192,6 +195,38 @@ pub fn generate_tables_for_indexer_sql(
         sql.push_str(&generate_internal_event_table_sql(&event_names, &schema_name, networks));
     }
 
+    if indexer.native_transfers.enabled {
+        let contract_name = NATIVE_TRANSFER_CONTRACT_NAME.to_string();
```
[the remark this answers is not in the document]
jaggad: Agree, I've added this as a check and error now.

core/src/indexer/start.rs
jaggad: I've extracted these out to two "sub-indexing" functions and run them concurrently 👍 

The only blocking operation in each was a single `await` to the `get_start_end_block` function, which would be in the order of a few hundred ms at most but even so I've run them concurrently now so should alleviate any additional latency this was likely to cause.

And when this is disabled it wont have any latency increase ofc

core/src/event/callback_registry.rs
jaggad: I'll extract it out to a function to share the logic, part of the challenge I had here (as I initially wanted to use 100% shared code for this) is backwards compatibility.

If I try to completely unify the `CallbackRegistry` and modify the type / input types to the `callback` itself, it will become a breaking change for any Rust bindings project types (nocode projects should be fine).

I opted to keep it non-breaking, so am limited in the refactor, but can definitely extract the duplicate code to a shared fn 👍

core/src/manifest/native_transfer.rs
jaggad: It's a little easier to work with if I can coalesce the two types into a single shared struct, rather than discriminating on the enum in multiple places.

This allows the core field to be like this (rather than typing to make `native_transfers: false` coalesce to `None`:
```rs
#[serde(default)]
#[serde(deserialize_with = "deserialize_native_transfers")]
pub native_transfers: NativeTransfers,
```

examples/rindexer_native_transfers/rindexer.yaml
jaggad: Ok, I've added logic to support this. NativeTransfer events will be auto opted in now if a stream is defined, but users can still override it by explicitly defining the event and aliasing it for example

core/src/manifest/native_transfer.rs
jaggad: Added

core/src/manifest/native_transfer.rs
jaggad: Chatbot doc examples added

core/src/event/callback_registry.rs
joshstevens19: i think if they just generated the new typings it would fix itself as its not a breaking change in the interface level right? aka rindexer codegen typings

core/src/manifest/native_transfer.rs
joshstevens19: does this means it always has to have enabled: true in the struct itself in the yaml or is it just hidden away?

jaggad: It's completely hidden away, the struct should only have `enabled: false` if someone wanted to define everything, but turn it off.. But i didn't document it because it's not intended to be used.

PR #171, opened 2025-03-26
Native transfer - trace method toggle

jaggad:
Alchemy is one of the few providers that does not support `trace_block` on L2s. However `trace_block` is better.

This PR enables configuration to switch to compatible `debug_traceBlockByNumber` instead if required due to provider limitations, however `trace_block` is still the default.

To investigate:
> Arbitrum had a fork & you need to use arbtrace_block before that block and then debug_traceBlockByNumber after

### DIFF
--- a/core/src/indexer/native_transfer.rs
+++ b/core/src/indexer/native_transfer.rs
@@ -139,7 +139,15 @@ pub async fn native_transfer_block_consumer(
     network_name: &str,
     config: &TraceProcessingConfig,
 ) -> Result<(), ProcessEventError> {
-    let trace_futures: Vec<_> = block_numbers.iter().map(|n| provider.trace_block(*n)).collect();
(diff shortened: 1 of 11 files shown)


core/src/indexer/native_transfer.rs
```diff
@@ -139,7 +139,15 @@ pub async fn native_transfer_block_consumer(
     network_name: &str,
     config: &TraceProcessingConfig,
 ) -> Result<(), ProcessEventError> {
-    let trace_futures: Vec<_> = block_numbers.iter().map(|n| provider.trace_block(*n)).collect();
+    let provider_call = async |block: U64| {
+        if 0.1 == 0.2 {
```
joshstevens19, line 143: 🤔

joshstevens19 (approved):

PR #197, opened 2025-05-22
Parallelize client creation

mikelsr:
We're testing rindexer with a large number of networks, and initialization can take a bit. We propose making it concurrent, which will make initialization faster in almost every multi-chain case. What do you think?

Thanks for your time!


joshstevens19:
0.17.0 this is now live thanks again for PR @mikelsr

PR #201, opened 2025-05-26
Fix Docker compatibility: Use x86-64-v2 target-cpu instead of native

r0zh:
## Problem
The current Dockerfiles uses `RUSTFLAGS='-C target-cpu=native'` which optimizes the Rust binary for the GitHub Actions runner's CPU architecture. This causes `Illegal instruction (core dumped)` errors when running the Docker images (`ghcr.io/joshstevens19/rindexer` and `ghcr.io/joshstevens19/rindexer-bundled`) on processors that don't support the same instruction sets as the build environment.

## Current Behavior
- ❌ Docker image crashes with "Illegal instruction (core dumped)" on various x86-64 processors (confirmed on Intel i7 10th gen)
- ❌ Binary is optimized for GitHub's build infrastructure rather than target deployment environments
- ❌ Limits portability across different CPU generations

## Proposed Solution
Change the RUSTFLAGS from:
```dockerfile
RUN RUSTFLAGS='-C target-cpu=native' ...
```

To:
```dockerfile
RUN RUSTFLAGS='-C target-cpu=x86-64-v2' ...
```

## Benefits
- ✅ Maintains good performance optimization (x86-64-v2 includes SSE4.2, POPCNT, etc.)
- ✅ Compatible with virtually all modern x86-64 processors (2009+)
- ✅ Prevents runtime crashes on older but still widely-used CPUs
- ✅ Better portability for Docker deployments across different environments

## Testing
Tested on Intel i7 10th gen where trying to use rindexer with the current image fails with "Illegal instruction (core dumped)" but making this change has resolved the issue.

## References
- [x86-64 microarchitecture levels](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels)
- [Rust target-cpu documentation](https://doc.rust-lang.org/rustc/codegen-options/index.html#target-cpu)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ RUN mkdir /app
 WORKDIR /app
 COPY . .
 RUN rustup target add x86_64-unknown-linux-musl
-RUN RUSTFLAGS='-C target-cpu=native' cargo build --release --target x86_64-unknown-linux-musl --features jemalloc
+RUN RUSTFLAGS='-C target-cpu=x86-64-v2' cargo build --release --target x86_64-unknown-linux-musl --features jemalloc
 
 FROM --platform=linux/amd64 scratch
 COPY --from=builder /app/target/x86_64-unknown-linux-musl/release/rindexer_cli /app/rindexer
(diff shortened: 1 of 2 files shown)


joshstevens19:
Thanks this is super useful will get it merged and release a quick one though if you install rindexer through the bash command ### 

“curl -L https://rindexer.xyz/install.sh | bash”  does it work for your i7 machine?

joshstevens19:
all live 

https://github.com/joshstevens19/rindexer/pkgs/container/rindexer-bundled/424722158?tag=sha-90e67aa432b31d8f109da0de0e15fa3945ed87a2 

https://github.com/joshstevens19/rindexer/pkgs/container/rindexer/424721754?tag=sha-90e67aa432b31d8f109da0de0e15fa3945ed87a2

thanks dude

PR #204, opened 2025-05-30
Support factory filter

desfero:


core/src/event/config.rs
```diff
@@ -15,14 +16,18 @@ use crate::{
...
 
-pub struct EventProcessingConfig {
+pub struct ContractEventProcessingConfig {
     pub id: String,
     pub project_path: PathBuf,
     pub indexer_name: String,
     pub contract_name: String,
     pub info_log_name: String,
-    pub topic_id: B256,
+    pub event_signature: B256,
```
desfero: TODO: revert this

core/Cargo.toml
```diff
@@ -67,6 +67,7 @@ tracing-subscriber = { version = "0.3", features = [
   "fmt",
   "time",
 ] }
+mini-moka = "0.10.3"
```
joshstevens19, line 70: whats this?

core/src/provider.rs
```diff
@@ -48,6 +48,10 @@ pub type RindexerProvider = FillProvider<
     RootProvider,
 >;
 
+// RPC providers have maximum supported addresses that can be provided in a filter
+// We play safe and limit to 100, in future maybe worth to make it provider specific for better performance
+const MAX_RPC_SUPPORTED_ACCOUNT_FILTERS: usize = 100;
```
joshstevens19: hmm limit is way to low i know alchemy do not have a max limit at all - which providers did you see max 100? feels like 1000 could be better but if you see others with lower? and if there is more it splits them and does many get logs right?

core/Cargo.toml
desfero: For in memory cache, we could go with `Mutex` or `RwLock` but that will lead to unnecessary locks.
This works in a way that lock is per `key` in `HashMap` which significantly lowers locks therefore increasing speed.

core/src/provider.rs
desfero: good point, double checked and alchemy can handle really a lot, was able to put 100k address and it still works,
tenderly has limit around 8k
not sure about others but makes sense to put 1000 as default
worth to also maybe do `in future maybe worth to make it provider specific for better performance` in case we detect alchemy, what do you think?

joshstevens19: yeah makes sense - maybe also an override in the yaml could also be good - this doesnt hurt performance just does many at the same time right?

desfero: done

documentation/docs/pages/docs/start-building/yaml-config/contracts.mdx
```diff
@@ -167,6 +168,62 @@ contracts: // [!code focus]
           - event_name: Approval // [!code focus]
 ```
 
+### factory
```
joshstevens19, line 171: would call out about the new setting we did here as a hint

joshstevens19 (approved):
epic PR 🔥

joshstevens19:
oh and changelog please

PR #209, opened 2025-06-10
Numeric U256 Array

jaggad:
Work in progress; not ready for review


core/src/database/postgres/sql_type_wrapper.rs
```diff
@@ -471,6 +474,35 @@ impl EthereumSqlTypeWrapper {
         out.extend_from_slice(&buf);
         Ok(IsNull::No)
     }
+
+    fn serialize_numeric_u256_array<T>(
```
joshstevens19, line 478: as its so close to serialize_numeric_array you think we could just make it share the method as only difference is value_converter signature and `write_u256_numeric_to_postgres` vs `write_numeric_to_postgres`

jaggad: I tried pretty hard to make it generic, but couldn't find a way, I will give it another shot and see but somehting about the U256 type being so different from normal makes it hard, can try again though.

joshstevens19: all good for now no worries

joshstevens19 (approved):

PR #214, opened 2025-06-14
added chain_id to TxInformation and to NetworkContract

### LINKED ISSUE #212: expose chain id in TxInformation
in the rindexer yaml you define networks like

```yaml
networks:
  - name: ethereum
    chain_id: 1
    rpc: ${ETHEREUM_RPC_URL}
```

when an event is emitted it also bakes in tx information which looks like:

```rs
#[derive(Debug, Serialize, Deserialize, Clone)]
pub struct TxInformation {
    pub network: String,
    pub address: Address,
    pub block_hash: BlockHash,
    pub block_number: U64,
    pub block_timestamp: Option<U256>,
    pub transaction_hash: TxHash,
    pub log_index: U256,
    pub transaction_index: U64,
}
```

We do not expose the chain_id here which is annoying as you may want to use it rather then your yaml name - we should expose it

PoulavBhowmick03:
Fixes #212 potentially


core/src/event/contract_setup.rs
```diff
@@ -30,6 +30,7 @@ pub struct NetworkContract {
     pub start_block: Option<U64>,
     pub end_block: Option<U64>,
     pub disable_logs_bloom_checks: bool,
+    pub chain_id: Option<u64>,
```
joshstevens19, line 33: most likely should not be Option as in theory its always known and required in the yaml - your going to have to also change the rust generation code as well here - https://github.com/joshstevens19/rindexer/blob/c72992ac19deb4259e8c26ff2e486f7bd72d2b5c/core/src/generator/events_bindings.rs

core/src/event/callback_registry.rs
```diff
@@ -83,6 +84,7 @@ impl EventResult {
             log: log.clone(),
             decoded_data: network_contract.decode_log(log.inner),
             tx_information: TxInformation {
+                chain_id: network_contract.chain_id.unwrap_or(0),
```
joshstevens19, line 87: should never not be known

core/src/event/callback_registry.rs
```diff
@@ -249,6 +251,7 @@ impl TraceResult {
             tx_information: TxInformation {
                 network: network.to_string(),
                 address: Address::ZERO,
+                chain_id: 0, // TODO: Chain ID should be set correctly if available.
```
joshstevens19, line 254: yeah should be set

rindexer_rust_playground/src/rindexer_lib/typings/rindexer_playground/events/erc_20_filter.rs
```diff
@@ -416,6 +416,7 @@ where
                 .map(|c| NetworkContract {
                     id: generate_random_id(10),
                     network: c.network.clone(),
+                    chain_id: None,
```
joshstevens19, line 419: https://github.com/joshstevens19/rindexer/blob/c72992ac19deb4259e8c26ff2e486f7bd72d2b5c/core/src/generator/events_bindings.rs this bit of code needs changing to cater for the chain id building for rust project

PoulavBhowmick03: got it

PR #216, opened 2025-06-17
Fix event dependency parsing

desfero:


core/src/indexer/dependency.rs
```diff
@@ -275,22 +266,48 @@ impl ContractEventDependencies {
...
+                        },
+                    });
+                }
+                None => {
+                    if let Some(dependency_event_tree) = contract_dependencies_tree {
+                        let dependency_tree = EventsDependencyTree::from_dependency_event_tree(
+                            &dependency_event_tree,
+                        );
+
+                        dependencies.push(ContractEventDependencies {
```
desfero, line 301: this was missing after refactoring, we need to populate dependencies array in case factory filter is not set

joshstevens19 (approved):
pending changelog

PR #221, opened 2025-06-23
Native index with `eth_getBlockByNumber`

jaggad:
1. Indexing with `eth_getBlockByNumber` for more efficiency (but still retain all debug/trace logic for future options)
2. Misc tweaks to optimise how this native transfer fetching is done with batching, and pass rpc provider to allow for better batching in other endpoints
3. Use correct provider base which is `AnyProvider` that can handle optimism, rollups, and all evm chain style responses.
4. Add a `Nullable` Numeric256 sql type
5. Add tx receipt endpoint as useful prep work for raw transaction indexing option (though, this isn't used now and could not be included if we want to exclude it)


core/src/indexer/native_transfer.rs
```diff
@@ -137,38 +140,183 @@ pub async fn native_transfer_block_fetch(
...
+    // Currently, `eth_getBlockByNumber` is a single JSON-RPC batch, and others are individual
+    // network calls so can be treated differently.
+    let (initial_concurrent_requests, limit_concurrent_requests) =
+        if is_rcp_batchable { (5, RPC_CHUNK_SIZE) } else { (5, 100) };
+
+    let mut concurrent_requests: usize = initial_concurrent_requests;
+    let mut buffer: Vec<U64> = Vec::with_capacity(limit_concurrent_requests);
+
+    loop {
+        sleep(Duration::from_millis(50)).await;
```
joshstevens19: should this not use the same setup we did with overrides in the logs one?

rindexer_rust_playground/src/rindexer_lib/indexers/rindexer_playground/playground_types_filter.rs
```diff
@@ -1,176 +1,94 @@
 #![allow(non_snake_case)]
-use std::{path::PathBuf, sync::Arc};
```
joshstevens19, line 2: files got messed up maybe just need to add formatter? we should generate the file for people nicely

rindexer_rust_playground/src/rindexer_lib/indexers/rindexer_playground/uniswap_v3_pool_filter.rs
```diff
@@ -1,151 +1,93 @@
 #![allow(non_snake_case)]
-use std::{path::PathBuf, sync::Arc};
```
joshstevens19, line 2: files got messed up maybe just need to add formatter? we should generate the file for people nicely

core/Cargo.toml
```diff
@@ -75,6 +75,7 @@ jemallocator = { version = "0.6.0", package = "tikv-jemallocator", optional = tr
 jemalloc-ctl = { version = "0.6.0", package = "tikv-jemalloc-ctl", optional = true }
 foundry-compilers = "0.17.0"
 alloy-chains = "0.2.2"
+log = "0.4.27"
```
0xSokka: Likely unwanted, common mistake.

core/src/event/callback_registry.rs
```diff
@@ -260,6 +262,38 @@ impl TraceResult {
...
+                address: Address::ZERO,
+                block_number: tx
+                    .block_number
+                    .map(U64::from)
+                    .expect("block_number should be present"),
+                block_timestamp: Some(U256::from(ts)),
+                transaction_hash: tx.tx_hash(),
+                block_hash: tx.block_hash.expect("block_hash should be present"),
+                transaction_index: U64::from(
+                    tx.transaction_index.expect("transaction_index should be present"),
```
0xSokka, line 290: Does the transaction index represent the nonce of a user or a transaction index in the block? if nonce then a U64 may be overkill.

core/src/indexer/fetch_logs.rs
```diff
@@ -48,13 +56,15 @@ pub fn fetch_logs_stream(
...
                 &max_block_range_limitation,
             ));
-            warn!(
-                "{}::{} - {} - max block range limitation of {} blocks applied - block range indexing will be slower then RPC providers supplying the optimal ranges - https://rindexer.xyz/docs/references/rpc-node-providers#rpc-node-providers",
-                config.info_log_name(),
-                config.network_contract().network,
-                IndexingEventProgressStatus::Syncing.log(),
-                max_block_range_limitation.unwrap()
-            );
+            if random_ratio(1, 50) {
```
0xSokka: What is the point of this? 1 in 50 chance of slowing down?

core/src/indexer/fetch_logs.rs
```diff
@@ -210,7 +221,16 @@ async fn fetch_historic_logs_stream(
             // clone here over the full logs way less overhead
             let last_log = logs.last().cloned();
 
-            if tx.send(Ok(FetchLogsResult { logs, from_block, to_block })).is_err() {
+            if tx.capacity() < 1 {
+                warn!(
```
0xSokka: This is good, will also help pinpoint issues and increase the permits. Good idea.

core/src/indexer/fetch_logs.rs
joshstevens19: to not give the warning constantly on the console

core/Cargo.toml
jaggad: what the hell, how did that get there 😆

0xSokka: Likely a typo when using `log` rather than `trace` happens to me often haha

core/src/event/callback_registry.rs
jaggad: Transaction index is within the block yes, u64 is probably overkill anyway but there's minimal diff betwren u32 and u64 for most operations and memory use

core/Cargo.toml
jaggad: Removed ty

core/src/indexer/fetch_logs.rs
jaggad: What josh said yes, otherwise it spams constantly

core/src/indexer/fetch_logs.rs
jaggad: Dropped the permit system now, it's more of an expected behaviour but does indicate producer is faster than the sink.

rindexer_rust_playground/src/rindexer_lib/indexers/rindexer_playground/uniswap_v3_pool_filter.rs
jaggad: Fixed this, had to move some things around as rustfmt couldn't parse the existing structure with an awaited nested callback

rindexer_rust_playground/src/rindexer_lib/indexers/rindexer_playground/playground_types_filter.rs
jaggad: Fixed

core/src/indexer/native_transfer.rs
jaggad: This is more to help not overload the http calls we make if we start getting responses too quickly, realistically we can probably remove it but feels better to keep for now.

unless i minsunderstand the question

core/src/database/postgres/generate.rs
```diff
@@ -151,7 +151,16 @@ fn generate_internal_event_table_sql(
             )
         }).collect::<Vec<_>>().join("\n");
 
-        format!("{}\n{}", create_table_query, insert_queries)
+        let create_latest_block_query = r#"CREATE TABLE IF NOT EXISTS rindexer_internal.latest_block ("network" TEXT PRIMARY KEY, "block" NUMERIC);"#.to_string();
```
jaggad, line 156: Tracking latest seen block in DB for external / more flexible progress update reporting

rindexer_rust_playground/abis/playground-types-abi.json
```diff
@@ -46,37 +46,37 @@
...
         "indexed": false,
-        "internalType": "int24",
+        "internalType": "int32",
         "name": "tick4",
         "type": "int32"
       },
       {
         "indexed": false,
-        "internalType": "int24",
+        "internalType": "int64",
```
jaggad, line 67: I don't think this was intentional as it's an incorrect abi and i think just handmade, so fixed this up as it was causing some silliness

core/src/database/postgres/client.rs
```diff
@@ -51,7 +51,7 @@ pub enum PostgresError {
     #[error("PgError {0}")]
     PgError(#[from] PgError),
 
-    #[error("Connection pool error: {0}")]
+    #[error("Postgres query error: {0}")]
```
joshstevens19: random rename?

core/src/database/postgres/generate.rs
joshstevens19: makes sense it at least allows you to know where rindexer is on the global level

core/src/indexer/fetch_logs.rs
```diff
@@ -164,7 +154,7 @@ async fn fetch_historic_logs_stream(
     );
 
     if from_block > to_block {
-        debug!(
+        error!(
```
joshstevens19: error?

core/src/indexer/fetch_logs.rs
```diff
@@ -197,20 +187,32 @@ async fn fetch_historic_logs_stream(
...
+                    IndexingEventProgressStatus::Syncing.log(),
+                    logs.len(),
+                    from_block,
+                    to_block
+                );
+            }
+
+            if tx.capacity() == 0 {
+                info!(
+                    "{}::{} - {} - Log channel full, indexer will backpressure (this is normal).",
```
joshstevens19: probs should make the log make more sense to a normal person - backpressure is not a term most people know

core/src/indexer/fetch_logs.rs
```diff
@@ -703,24 +728,51 @@ fn retry_with_block_range(
...
+        return Some(RetryWithBlockRangeResult {
+            from: from_block,
+            to: halved_to_block,
+            max_block_range: None,
+        });
+    }
+
+    // We can't keep up with our own sending rate. This is rare, but we must backoff throughput.
+    if error_message.contains("error sending request") {
+        tokio::time::sleep(Duration::from_secs(3)).await;
```
joshstevens19: 3 second a magic number?

core/src/indexer/last_synced.rs
```diff
@@ -246,7 +247,32 @@ pub fn update_progress_and_last_synced_task(
                 .await;
 
             if let Err(e) = result {
-                error!("Error updating last synced block: {:?}", e);
+                error!("Error updating db last synced block: {:?}", e);
+            }
+
+            let latest = config
```
joshstevens19: we already have a db command above running not better to do this in 1 execute over 2 db ones?

core/src/indexer/native_transfer.rs
```diff
@@ -137,38 +140,182 @@ pub async fn native_transfer_block_fetch(
...
+            let is_rate_limit_error = matches!(&e, ProcessEventError::ProviderCallError(
+                ProviderError::RequestFailed(
+                    RpcError::Transport(
+                        TransportErrorKind::HttpError(http_err)
+                    )
+                )) if http_err.status == 429
+            );
+
+            if is_rate_limit_error {
+                error!(
```
joshstevens19: is this an error feels just like a warning..

core/src/indexer/process.rs
```diff
@@ -561,19 +547,17 @@ async fn handle_logs_result(
                 tokio::time::sleep(Duration::from_millis(1000)).await;
             }
 
-            if !fn_data.is_empty() {
-                return if config.index_event_in_order() {
+            // Important that we call this for every event even if there are no logs.
+            // This is because we need to sync the last seen block
+            if config.index_event_in_order() {
+                trigger_event(config, fn_data, result.to_block).await;
```
joshstevens19, line 558: we did this on purpose we did not want to have to write a db event up if nothing has changed to overload the db each time - we have a global last seen which is good but i dont think we want to trigger an event with no logs at all?

rindexer_rust_playground/src/rindexer_lib/typings/rindexer_playground/mod.rs
```diff
@@ -1,7 +1,2 @@
 #![allow(dead_code, unused)]
-
-/// THIS IS A GENERATED FILE. DO NOT MODIFY MANUALLY.
```
joshstevens19, line 3: how come we lose this being wrote to the files?

joshstevens19 (changes requested):
overall looks good my only concern is the firing of the trigger event even with empty logs we wouldnt want callbacks to be hit on empty logs - i understand we fired to last seen update but i think we have to work out how to not fire the main event as well

core/src/database/postgres/client.rs
jaggad: Reverted

core/src/indexer/fetch_logs.rs
jaggad: Changed to `warn` as it self recovers

core/src/indexer/fetch_logs.rs
jaggad: Changed to "Log channel full, waiting for events to be processed"

core/src/indexer/fetch_logs.rs
jaggad: Yep, just a random backoff rate, dropped it to `1` for now.

core/src/indexer/native_transfer.rs
jaggad: Yeah, warn is better.

rindexer_rust_playground/src/rindexer_lib/typings/rindexer_playground/mod.rs
jaggad: No idea, just regenerated and nothing else was changed so maybe a change many versions ago?

core/src/indexer/process.rs
jaggad: Added the length check to the actual handler trigger now, rather than at this level which is the incorrect level.

We need to persist last seen block even if there were no logs in that block range, but i now don't fire that handler if there's nothing (just the db writes)

core/src/indexer/last_synced.rs
jaggad: Batch executing now in a simple query pipeline.

PR #224, opened 2025-07-03
Index with eth_getBlockByNumber, Overhaul Permit System, Memory Optimise, Fix Codegen

jaggad:
Major fixes and system overhaul

1. Indexing with `eth_getBlockByNumber` for more efficiency (but still retain all debug/trace logic for future options)
2. Misc tweaks to optimise how this native transfer fetching is done with batching, and pass rpc provider to allow for better batching in other endpoints
3. Use correct provider base which is `AnyProvider` that can handle optimism, rollups, and all evm chain style responses.
4. Add a `Nullable` Numeric256 sql type
5. Add tx receipt endpoint as useful prep work for raw transaction indexing option (though, this isn't used now and could not be included if we want to exclude it)
6. Remove the permits system entirely, it's had some serious problems with fair distribution (where some events would consistenyl take more permits than others)
7. Add an ever-so-slightly backpressured queue per "network-event" for super fair scheduling and decreased write contention on the database
8. We weren't writing to the db if no events were found in a block range... this was a major problem for highly infrequent events as rindexer wouldn't write the "last seen block", fixed.
9. Memory on large-scale indexer before was peaking and OOMing at 40gb+, now it is optimally sitting at 12.5gb
10. Fixed a pretty major bug with the "optimal log parsing regex" (where the BlockNumber::from_str() wasn't actually parsing the hex, so it silently failed and used sub-optimal fallbacks)

I also added bounded concurrency to the handler trigger / callback along with documentation which we can merge into this one if we want to: https://github.com/jaggad/rindexer/pull/10/files


core/src/database/postgres/generate.rs
```diff
@@ -134,24 +134,35 @@ fn generate_internal_event_table_sql(
...
     schema_name: &str,
     networks: Vec<&str>,
+    start_blocks: Vec<Option<U64>>,
 ) -> String {
     abi_inputs.iter().map(|event_info| {
         let table_name = generate_internal_event_table_name(schema_name, &event_info.name);
 
         let create_table_query = format!(
-            r#"CREATE TABLE IF NOT EXISTS rindexer_internal.{} ("network" TEXT PRIMARY KEY, "last_synced_block" NUMERIC);"#,
+            r#"CREATE TABLE IF NOT EXISTS rindexer_internal.{} ("network" TEXT PRIMARY KEY, "start_block" NUMERIC, "last_synced_block" NUMERIC);"#,
```
joshstevens19: this is a breaking change to existing indexers - i had a plan now to handle breaking changes we should expose a `rindexer migrate <version>` which then updates the tables for them OR start storing the rindexer version in DB so we can auto migrate them on the start (similar to flyway) - for example on lens this would be nightmare now and we really now do need to keep in mind breaking changes hurt things when we running this in production fully now so we need to work it out before we merge this

core/src/database/postgres/generate.rs
```diff
@@ -134,24 +134,35 @@ fn generate_internal_event_table_sql(
...
             table_name
         );
 
-        let insert_queries = networks.iter().map(|network| {
+        let insert_queries = networks.iter().zip(start_blocks.clone()).map(|(network, start_block)| {
             format!(
-                r#"INSERT INTO rindexer_internal.{} ("network", "last_synced_block") VALUES ('{}', 0) ON CONFLICT ("network") DO NOTHING;"#,
+                r#"INSERT INTO rindexer_internal.{} ("network", "start_block", "last_synced_block") VALUES ('{}', '{}', 0) ON CONFLICT ("network") DO NOTHING;"#,
```
joshstevens19: what does start block mean? why do we need it in the internal table? you either start from 0 (aka you dont define it in yaml) or you define it in yaml itself.. what benefit is this having ? as the breaking change is way worse here so any nice to haves has to be really justified

core/src/event/callback_registry.rs
```diff
@@ -120,7 +122,7 @@ pub struct EventCallbackRegistryInformation {
 
 impl EventCallbackRegistryInformation {
     pub fn info_log_name(&self) -> String {
-        format!("{}::{}", self.contract.name, self.event_name)
+        format!("{}::{}", self.contract.name.replace("Filter", ""), self.event_name)
```
joshstevens19: do you remember what this fixed ?

core/src/generator/events_bindings.rs
```diff
@@ -833,22 +837,38 @@ pub fn generate_event_handlers(
...
+                .to_string();
 
             for item in &abi_name_properties {
                 if let Some(wrapper) = &item.ethereum_sql_type_wrapper {
                     data.push_str(&format!(
-                        "EthereumSqlTypeWrapper::{}(result.event_data.{}{}),",
+                        "\nEthereumSqlTypeWrapper::{}({}result.event_data.{}{}),",
                         wrapper.raw_name(),
-                        item.value,
+                        match wrapper {
```
joshstevens19, line 848: context on why this part changed?

core/src/provider.rs
```diff
@@ -50,17 +52,27 @@ pub type RindexerProvider = FillProvider<
...
+    RootProvider<AnyNetwork>,
+    AnyNetwork,
 >;
 
 // RPC providers have maximum supported addresses that can be provided in a filter
 // We play safe and limit to 5000 by default, but that can be overridden in the configuration
 const DEFAULT_RPC_SUPPORTED_ACCOUNT_FILTERS: usize = 5000;
 
+/// Maximum RPC batching size available for the provider.
+pub const RPC_CHUNK_SIZE: usize = 1000;
```
joshstevens19, line 65: if running this on a tenderly public node does it still work just slower?

rindexer_rust_playground/src/rindexer_lib/indexers/rindexer_playground/playground_types_filter.rs
```diff
@@ -36,42 +35,45 @@ async fn swap_handler(manifest_path: &PathBuf, registry: &mut EventCallbackRegis
...
-                    result.tx_information.log_index.to_string()
+                    result.tx_information.log_index.to_string(),
                 ]);
                 let data = vec![
                     EthereumSqlTypeWrapper::Address(result.tx_information.address),
                     EthereumSqlTypeWrapper::Address(result.event_data.sender),
                     EthereumSqlTypeWrapper::Address(result.event_data.recipient),
-                    EthereumSqlTypeWrapper::I256(result.event_data.amount0),
-                    EthereumSqlTypeWrapper::I256(result.event_data.amount1),
+                    EthereumSqlTypeWrapper::I256(I256::from(result.event_data.amount0)),
```
joshstevens19, line 49: as amount0 is a int256 why does it need to wrap here? same for all the I256::from - its a very odd syntax to promote and if we can abstract it away so they just use the EthereumSqlTypeWrapper::I256 its much more nice for the developer

core/src/event/callback_registry.rs
jaggad: This wasn't actually a fix, it just reduced the length of the rindexer log lines and made it a bit easier to read by removing `Filter` from everything in the logging.

core/src/database/postgres/generate.rs
jaggad: That's a good point actually, I think we could most easily address this by creating with the column, or adding it if it doesn't exist in a separate few queries.

But this was to add better support for external tracking of progress by exposing the `last, current, and first` blocks seen per network-event in the database.

This means we can have more flexible progress reporting and checks going on. We can technically get by without the breaking change here since we can grab it from the yaml instead but this seemed like a simple solution and a decent way to get more data available from the DB for later (if we build dashboards, etc)

core/src/database/postgres/generate.rs
jaggad: Start block is a DB synced value of what the yaml defined, answered in above context, maybe best to discuss in that comment thread

rindexer_rust_playground/src/rindexer_lib/indexers/rindexer_playground/playground_types_filter.rs
jaggad: You're actually right, `amount0` here doesnt' need to be using `I256::from`, that's an oversight i think.

However some of the other other H192, H160 fields, etc need to be converted to a compatible U256 or I256 type. So either we make `I160`, etc. fields (there are many more we'd want to support), or we promote the `X::from` pattern.

Probably we'd be better off making the `EthereumSqlTypeWrapper` more generic, so a `U256` can accept any parameter that can be converted to a `U256` rather than just a U256 for example. Basically design to traits rather than to concrete types.

What do you think?

core/src/generator/events_bindings.rs
jaggad: Was related to some missed parts in the alloy upgrade. We don't have H160, H192, etc support in the `EthereumSqlWrapper` and alloy now treats these fields as the actual bit types rather than remapping to a U256 for example.

This means we need to support / do this mapping ourselves. The two options were

1. Extend `EthereumSqlWrapper` with many more alloy types, a fine possibility
2. Use the `From` trait to convert them in the handler themselves

I think overall we may actually want to rethink the `EthereumSqlWrapper` to be a bit more ergonomic, for example being more of a glue between PG types and Alloy types and using more traits and generic implementations to support both `u64 and U64` for example since alloy now changed to use `u64` whilst some rindexer internals use `U64`.

Just an idea on how we could make this better, the `match` itself was added to address incompatible types.

core/src/provider.rs
jaggad: As far as I can tell it's just Alchemy that enforces the `1000` max batch size, that being said, it's a reasonably upper limit as beyond or even at this point any single slow call response will slow the whole batch reducing concurrency and throuhgput.

From my testing and from recommendations of most providers, keeping it under this limit is preferrable

core/src/database/postgres/generate.rs
jaggad: > overall great PR just need to address the breaking change really and if it was for just more visibility we should omit it for now as it will hurt production upgrades

But yeah this is it, it's really just for visibility.

joshstevens19: yeah we should remove because its a critical breaking change

core/src/event/callback_registry.rs
joshstevens19: i think this was in there for a reason so i dont think we should remove as we do some magic on filter term

core/src/generator/events_bindings.rs
joshstevens19: i would most likely just extend the EthereumSqlWrapper to have these types as it is growing we do need to think longer term but end of day it maps directly to type so easy for dev to workout imho

rindexer_rust_playground/src/rindexer_lib/indexers/rindexer_playground/playground_types_filter.rs
joshstevens19: i think its fine to do the .into() if needed on the more types like I160 but then again im also not that bothered in extending the EthereumSqlTypeWrapper to be inline as then it is so easy to explain based on the types itself

rindexer_rust_playground/src/rindexer_lib/indexers/rindexer_playground/playground_types_filter.rs
```diff
@@ -36,42 +35,45 @@ async fn swap_handler(manifest_path: &PathBuf, registry: &mut EventCallbackRegis
...
-                    EthereumSqlTypeWrapper::I32(result.event_data.tick4.as_i32()),
-                    EthereumSqlTypeWrapper::I64(result.event_data.tick5.as_i64()),
-                    EthereumSqlTypeWrapper::I128(i128::from_be_bytes(result.event_data.tick6.to_be_bytes())),
-                    EthereumSqlTypeWrapper::I256(I256::from_dec_str(&result.event_data.tick7.to_dec_string()).expect("I24 fits in I256")),
+                    EthereumSqlTypeWrapper::I8(result.event_data.tick2),
+                    EthereumSqlTypeWrapper::I16(result.event_data.tick3),
+                    EthereumSqlTypeWrapper::I32(result.event_data.tick4),
+                    EthereumSqlTypeWrapper::I64(result.event_data.tick5),
+                    EthereumSqlTypeWrapper::I128(result.event_data.tick6),
+                    EthereumSqlTypeWrapper::I256(I256::from(result.event_data.tick7)),
```
joshstevens19, line 59: int192 is odd i guess this is much better then I256::from_dec_str(&result.event_data.tick7.to_dec_string()).expect("I24 fits in I256") lol but i dont know something in my mind thinks EthereumSqlTypeWrapper::I192(result.event_data.tick7) is easier for the dev to auto-complete even if that means EthereumSqlTypeWrapper is full of types - maybe a more generic one of 

Could EthereumSqlTypeWrapper::to_sql<T>(T) work as a generic catch-all?

jaggad: I think `EthereumSqlTypeWrapper::from(T)` could work as a generic catch all yes, we just need to implement `From<T>` and `TryFrom<T>` traits for all of the specific. types that are supported by alloy and can simplify this with macros to reduce boilerplate as needed.

I still think the concrete types are good though for fine tuning, but it's more related to postgres types than eth types and we can just support a `EthereumSqlTypeWrapper::from(result.event_data.tick6)` as a default impl of most alloy types to simplify things if we want to make it prettier.

It all still needs the specific `I` and `U` types though. 

Maybe in a follow up PR we can address this, since it's probbaly a bit out of scope of this PR? I just initially wanted to fix the failing `codegen indexer` compilation as a part of this PR.

core/src/event/callback_registry.rs
jaggad: This is just the log line itself, so it's non breaking for sure. But I removed it now.

joshstevens19 (approved):

PR #235, opened 2025-07-15
Block Clock, Slight Refactor, Dependency Upgrades

jaggad:
# Block Clock

------
- Add block clock
- Update documentation
- Cleanup deps a little
- Additionally refactor rust playground into examples, and helm into providers
------

Allow users to efficiently attach timestamps to logs.

We can add further optimisations to pre-compute large ranges of delta-run-encoded timestamps which for most chains will allow us to expose a "megabyte scale" file to accurately calculate block timestamps.

This would be considered an optimisation and this base work of fetching timestamps if they are not present in logs represents the effective baseline.

Follow up PR (Handler for raw block/txs with native tracing): https://github.com/jaggad/rindexer/pull/15

### DIFF
--- a/documentation/docs/pages/docs/changelog.mdx
+++ b/documentation/docs/pages/docs/changelog.mdx
@@ -5,10 +5,11 @@
 
 ### Features
 -------------------------------------------------
-- fix: descope some logs
 - Adds timestamp config to the yaml config, allowing users to opt-in to block timestamps in logs
 - Add chain_id to TxInformation struct
-- Add xtasx with block timestamp encoding mechanism
+- Add xtask with block timestamp encoding mechanism
+- Adds nocode postgres migration "versioning system"
+- Expose a rust project handler to process raw blocks with transactions if native indexing is enabled.
 
 ### Bug fixes
 -------------------------------------------------


joshstevens19 (changes requested):
```
- enable_timestamp: true

- config:
    timestamps: # override it if you want
      enabled: true // [!code focus]
      sample_rate: 0.25 // [!code focus

  include_events:
    - name: Transfer
      enable_timestamp: true
    - Approval
```

core/src/blockclock/fixed.rs
```diff
@@ -0,0 +1,185 @@
+//
+// base (8453)
```
joshstevens19: this needed?

core/src/blockclock/fixed.rs
```diff
@@ -0,0 +1,185 @@
...
+#[derive(Debug)]
+pub enum SpacedNetwork {
+    Base(Chain),
+    Blast(Chain),
+    Soneium(Chain),
+    Worldchain(Chain),
+}
+impl SpacedNetwork {
+    /// The genesis unix timestamp for the network (the zero block).
+    pub fn genesis_time(&self) -> i64 {
```
joshstevens19: any reason this is signed? would say genesis_time can not be minus, is it due to postgres?

core/src/blockclock/fixed.rs
```diff
@@ -0,0 +1,185 @@
...
+        match &self {
```
joshstevens19, line 17: we could make this automatic by just getting the block 0?

core/src/blockclock/fixed.rs
```diff
@@ -0,0 +1,185 @@
...
+            SpacedNetwork::Base(_) => 1686789347,
+            SpacedNetwork::Blast(_) => 1708809815,
+            SpacedNetwork::Soneium(_) => 1733134751,
+            SpacedNetwork::Worldchain(_) => 1719335639,
+        }
+    }
+
+    /// Get the blocktime for a chain if present
+    pub fn inner(&self) -> &Chain {
```
joshstevens19, line 26: we should try to do as many as we can here

core/src/blockclock/fixed.rs
```diff
@@ -0,0 +1,185 @@
...
+impl TryFrom<&Chain> for SpacedNetwork {
+    type Error = String;
+
+    fn try_from(value: &Chain) -> Result<Self, Self::Error> {
+        match value.named() {
+            Some(NamedChain::Base) => Ok(SpacedNetwork::Base(value.to_owned())),
+            Some(NamedChain::Blast) => Ok(SpacedNetwork::Blast(value.to_owned())),
+            Some(NamedChain::Soneium) => Ok(SpacedNetwork::Soneium(value.to_owned())),
+            Some(NamedChain::World) => Ok(SpacedNetwork::Worldchain(value.to_owned())),
+            _ => Err(format!("{:?} is not a spaced network", value)),
```
joshstevens19, line 75: 1.88 stable allows you to do format!("{value} is not a spaced network")

core/src/blockclock/fixed.rs
```diff
@@ -0,0 +1,185 @@
...
+            network.inner().id(),
+            None,
+            None,
+            None,
+            Default::default(),
+            None,
+            None,
+        )
+        .await
+        .unwrap()
```
joshstevens19, line 119: probs can make check_block_times a result and bubble this back up

core/src/blockclock/fixed.rs
```diff
@@ -0,0 +1,185 @@
...
+        .get_block_by_number_batch(&blocks_req, false)
+        .await
+        .unwrap()
```
joshstevens19, line 122: same here

core/src/blockclock/fixed.rs
```diff
@@ -0,0 +1,185 @@
...
+        .into_iter()
+        .map(|t| (t.header.number, t.header.timestamp as i64))
+        .collect::<HashMap<_, _>>();
+
+        for (k, v) in block_time {
```
joshstevens19, line 127: probs no speed increase here (unless get_block_time does anything special)  but can probs fire network.get_block_time(k) togetehr in one with no loop?

core/src/blockclock/mod.rs
```diff
@@ -0,0 +1,367 @@
...
+            return dates;
+        }
+
+        anchors.sort_by(|(block_a, _), (block_b, _)| block_a.partial_cmp(block_b).unwrap());
+
+        let (first, _) = *anchors.first().unwrap();
+        let (last, _) = *anchors.last().unwrap();
+        let blocks = first..=last;
+
+        for block in blocks.clone() {
```
joshstevens19: not hard on it but could be a map

core/src/event/callback_registry.rs
```diff
@@ -42,6 +42,7 @@ pub enum CallbackResult {
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct TxInformation {
+    pub chain_id: u64,
```
joshstevens19, line 45: nice

core/src/event/callback_registry.rs
```diff
@@ -85,6 +86,7 @@ impl EventResult {
             log: log.clone(),
             decoded_data: network_contract.decode_log(log.inner),
             tx_information: TxInformation {
+                chain_id: network_contract.cached_provider.chain.id(),
```
joshstevens19, line 89: rust project will need changing to add this also

documentation/docs/pages/docs/start-building/yaml-config/top-level-fields.mdx
```diff
@@ -88,4 +88,27 @@ repository: https://github.com/joshstevens19/rindexer
 config:
   buffer: 2 // [!code focus]
   callback_concurrency: 4 // [!code focus]
+```
+
+## timestamps
```
joshstevens19, line 93: timestamps > enable_timestamps_for_all: bool (or whatever)

examples/rindexer_rust_playground/src/rindexer_lib/typings/rindexer_playground/events/erc_20_filter.rs
```diff
@@ -145,9 +146,9 @@ where
         Fut: Future<Output = EventCallbackResult<()>> + Send + 'static,
     {
         let csv = AsyncCsvAppender::new(
-            r"/Users/skanda/Projects/rindexer/cli/../rindexer_rust_playground/generated_csv/ERC20Filter/erc20filter-approval.csv",
+            r"/Users/jackedgson/Development/avara/rindexer/examples/rindexer_rust_playground/generated_csv/ERC20Filter/erc20filter-approval.csv",
```
joshstevens19, line 149: we should fix this to read from location lol

core/src/blockclock/fixed.rs
jaggad: Yeah, can we turn off the disabled lint and run this formatting on the whole project. I would prefer to inline but as long as we have the lint disabled i will likely keep forgetting

examples/rindexer_rust_playground/src/rindexer_lib/typings/rindexer_playground/events/erc_20_filter.rs
jaggad: Haha for sure, will see if i can bundle it tomorrow

core/src/blockclock/fixed.rs
jaggad: Nope, leftover.

core/src/blockclock/fixed.rs
jaggad: Just got used to representing it as i64 due to postgres but it can be a u64. Changed.

core/src/blockclock/fixed.rs
jaggad: We could yes, I think this is fine enough that as it's a hard-per-chain opt-in in the code here so not currently something than can be made generic.

core/src/blockclock/fixed.rs
jaggad: I've tested many others, I would like to circle back to this in the future when I do the delta-run-length encoded timestamp setup for a bunch of popular chains first.

core/src/blockclock/fixed.rs
jaggad: It's just a test case, I dont think it matters and unwrapping is fine

core/src/blockclock/fixed.rs
jaggad: I'm ok to unwrap in a `test`

core/src/blockclock/fixed.rs
jaggad: This is a unit test, not worth the change.

core/src/blockclock/mod.rs
jaggad: What could be a map sorry? You mean a hashmap rather than a btree map?

core/src/event/callback_registry.rs
jaggad: Has been added everywhere it was needed and it works fine, i dont think anything else is needed

core/src/blockclock/fixed.rs
joshstevens19: why would we keep it manual when its so easy to make it just read from rpc ?

core/src/blockclock/mod.rs
joshstevens19: think its fine to leave

core/src/blockclock/fixed.rs
joshstevens19: my bad didnt see it was a test

core/src/blockclock/fixed.rs
joshstevens19: my bad didnt see it was a test

core/src/event/callback_registry.rs
joshstevens19: awesome

joshstevens19 (changes requested):
one other thing i thought about last night here is the no-code creates the tables for the developer, it should create the timestamp row now as well and write to it - alongside thats a breaking change so we should handle the migration for them? else it would break when they upgrade on no-code.. was thinking about this a bit and maybe we should start writing the version number it was last ran on somewhere so we can start migrating for the user over them having to run scripts / start from nothing

core/src/blockclock/fixed.rs
jaggad: RPC don't all return accurate genesis block timestamps from what I can tell, at least there is a difference between what blockscanner like `etherscan` return and what RPCs return "0x0".

A genesis time of `0` would not work with the current `SpacedNetwork` implementation, and I'd need to deprecate it in favor of the delta-encoded timestamps everywhere if we want the zero block to actually be 1970.. which I would argue is somewhat unexpected behavior from the perspective of a user.

xtask/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "xtask"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+clap = { version = "4.5.4", features = ["derive", "env"] }
+anyhow = { version = "1.0.98" }
+tokio = { version = "1.46.1", features = ["full"] }
+tracing = { version = "0.1.41" }
+tracing-subscriber = { version = "0.3.19", features = [
```
joshstevens19, line 11: whats xtask?

joshstevens19 (approved):

xtask/Cargo.toml
jaggad: its a handy script runner (https://github.com/matklad/cargo-xtask) that allows me to write nice little cli utils that should only be available to pacakge devs and we can run them like so (e.g to update blockclocks with more recent blocks):
```
cargo xtask update-block-clocks --alchemy-api-key XXX --batch-size 1000
```

PR #261, opened 2025-07-15
feat: feature gate reth

SkandaBhat:


joshstevens19 (changes requested):
as said on telegram we need to find all reference to cargo build and make sure to include the feature so it compiles in no-code binary

PR #270, opened 2025-07-20
feat: update GraphQL server build process

### LINKED ISSUE #267: Build GraphQL server from source in `build.rs` instead of committing binaries
**Problem:**

- The current approach of committing pre-compiled binaries for the GraphQL server (`rindexer-graphql-linux`, `rindexer-graphql-macos`, `rindexer-graphql-win.exe`) introduces significant security risks, maintenance burdens, and portability issues. 
- The `graphql/package.json` reveals these binaries are built with Node.js 16, which is **end-of-life** and no longer receives security updates. 
- Binaries significantly increase the size of the Git repository, making it slower to clone and manage.
- Binaries are likely compiled for a specific CPU architecture (e.g., x86_64). They will fail to run on other architectures, such as ARM

**Proposed solution:**

- Remove Binaries from Git
- Update Node.js and dependencies
- Enhance `core/build.rs`: The build should target only the host OS and architecture (e.g., node20-linux-x64 on Linux) to speed up the build

isSerge:
Closes #267 

Main changes:

1. Remove binaries from repository
2. Update Node.js and dependencies in `package.json`
3. Update `core/build.rs`:

- utilize `@yao-pkg/pkg` to build binaries (original `@vercel/pkg` has been archived and this is a community-maintained fork)
- build binary for host OS and architecture, including both `x86_64` and `aarch64`
- add re-build logic

4. Simplify executable path retrieval in `api/graphql.rs` by utilizing `RINDEXER_GRAPHQL_EXE` temporary env variable

I have tested on `macos-arm64`, would be great if we could verify it works on other platforms, especially Windows


core/build.rs
```diff
@@ -1,26 +1,230 @@
...
+    os: String,
+    arch: String,
+    exe_name: String,
+}
+
+fn get_target_info() -> TargetInfo {
+    let os = env::var("CARGO_CFG_TARGET_OS").expect("CARGO_CFG_TARGET_OS not set");
+    let arch = env::var("CARGO_CFG_TARGET_ARCH").expect("CARGO_CFG_TARGET_ARCH not set");
+
+    let node_arch = match arch.as_str() {
```
joshstevens19, line 112: i think it may be worth having a github action which installs the binary and runs it and sees if it works? im not sure how else to test this on all operating systems and its a pretty risky merge

joshstevens19 (changes requested):
this is a really good PR but i dont think i can merge this until we have a github action which simulates it works even if we just get to see it running so we can be happy it works everywhere

core/build.rs
isSerge: we can add a smoke test right after the build

joshstevens19: yes i think it would be good just because i am in same boat only have a mac and the CI has everything so if we can have smoke tests of it actually being able to run it would be good - alongside worth checking https://github.com/joshstevens19/rindexer/blob/master/documentation/docs/public/install.sh this script now as i think if its in binary we can remove the resources part of it? https://github.com/joshstevens19/rindexer/blob/master/documentation/docs/public/install.sh#L80

joshstevens19:
i go on holiday shortly ive assigned @desfero to review this and test it manually to make sure it works before we merge it

isSerge:
@joshstevens19 @desfero smoke tests added

desfero (approved):
LGTM, let's merge it with latest master which has clippy lints fixed

isSerge:
my branch is actually in sync with `master`, not sure why clippy is not happy

isSerge:
@joshstevens19 do we still want two dockerfiles or a single one (with Node.js included)?

joshstevens19:
@isSerge seems we missed some places in the release ci - https://github.com/joshstevens19/rindexer/actions/runs/17077833090/job/48423783638 i guess i can remove any reference of uploading this now because its in binary itself?

isSerge:
@joshstevens19 no, this is updated CI and we still need a binary, the problem is it cannot find the binary for some reason

isSerge:
@joshstevens19 I would suggest reverting your recent fix, I can investigate why it could not find binary for MacOS (likely generated binary name mismatch), but probably tomorrow - it's already end of the day in my timezone

joshstevens19:
I think I’ve got a fix will let you know trying again @isSerge

PR #273, opened 2025-07-22
Update install.sh for x86_64

### LINKED ISSUE #272: On Mac x86_64 install.sh fails to download
When running the install scripts: sudo curl -L https://rindexer.xyz/install.sh | bash

Get the following error:

Installing rindexer version 0.21.2...
Preparing the installation...
Downloading binary archive from https://github.com/joshstevens19/rindexer/releases/download/v0.21.2/rindexer_darwin-x86_64.tar.gz...
curl: (56) The requested URL returned error: 404
Failed to download rindexer version 0.21.2. Please check if the version exists.
Available releases: https://github.com/joshstevens19/rindexer/releases

Also tried to run the version specific install for v0.21.2, v0.21.1 and v0.21.0 and got the same 404. ChatGPT suggested modifying the install script, will try to do a PR with the change that got it to install for me.

XanDev3:
closes #272 

This is what worked for me to install, if there was some config I missed then maybe documentation needs to be updated instead and then feel free to disregard my issue. Not sure where this was intended to be handled in the original script as the comment suggests.


joshstevens19 (approved):
great find thanks

PR #276, opened 2025-07-25
Generate proper snake case file names for factory contracts handlers

desfero:


core/src/manifest/core.rs
```diff
@@ -122,7 +123,7 @@ impl Manifest {
                     }
 
                     // suffix with factory filter details to allow having the same contract name at the `contracts` level in yaml
-                    let overridden_factory_contract_name = format!("{}{}{}", first_factory.name, first_factory.event_name, first_factory.input_name);
+                    let overridden_factory_contract_name = format!("{}{}{}", first_factory.name, to_pascal_case(&first_factory.event_name), to_pascal_case(&first_factory.input_name));
```
desfero, line 126: making `event_name` also pascal case given on lens we name events like: `Lens_NamespaceFactory_Deployment`

PR #285, opened 2025-08-11
Support tuples and nested tuples in no code and rust

desfero:


core/src/generator/events_bindings.rs
```diff
@@ -712,6 +713,36 @@ pub fn generate_event_bindings(
         .map_err(GenerateEventBindingsError::GenerateEventBindingCode)
 }
 
+pub fn generate_event_input_path(property: &AbiProperty) -> (String, bool) {
```
desfero, line 716: TODO: Add some comments or even unit tests

core/src/abi.rs
```diff
@@ -82,22 +109,33 @@ impl ABIInput {
     pub fn generate_abi_name_properties(
         inputs: &[ABIInput],
         properties_type: &GenerateAbiPropertiesType,
-        prefix: Option<&str>,
-    ) -> Vec<GenerateAbiNamePropertiesResult> {
+        path: &Option<Vec<AbiNamePropertiesPath>>,
```
jaggad: `nitpick`: Slightly non-idomatic rust to pass a reference to an option, typically the option would simple passed plainly with an inner type reference, perhaps you could try this (or a vec if needed).

```suggestion
        path: Option<&[AbiNamePropertiesPath]>
```

core/src/database/postgres/sql_type_wrapper.rs
```diff
@@ -1293,32 +1286,45 @@ fn map_dynamic_uint_to_ethereum_sql_type_wrapper(
 fn map_log_token_to_ethereum_wrapper(
     abi_input: &ABIInput,
     token: &DynSolValue,
-) -> EthereumSqlTypeWrapper {
+) -> Vec<EthereumSqlTypeWrapper> {
```
jaggad, line 1289: It's unlikely to be the bottleneck for this kind of an application but putting everything in a vec does mean more heap allocations as far as I understand, just something to be mindful of, but i see the need for a list type for the tuple so seems fine enough to me given the functionality gain

jaggad (commented):
Not done with review, will do more tomorrow but just wanted to provide what I have (which is really only nitpicks for now).

Overall i dont have any major feedback so far

core/src/abi.rs
desfero: fixed

core/src/generator/events_bindings.rs
desfero: added tests

Issue #286, opened 2025-08-15
[Feature request] Extended events filtering


serzhiio:
### **Summary**
Index all ERC-20 Transfer events related to each pool that is discovered from a factory’s PoolCreated events—in a single table, without duplication—and be able to filter those transfers by source/destination address (e.g., only rows where from==pool or to==pool). Today I can subscribe to tokens by address, but I can’t easily constrain results to transfers involving the pool, nor merge token0+token1 streams into one sink table cleanly.

### **Motivation / Use case**
For Uniswap/HyperSwap-like DEXes, each pool has two tokens (token0/token1) and a pool address. 
TVL/flow analytics require:
- listing all transfers where the pool is from (withdrawal) or to (deposit),
- across both token0 and token1,
- deduplicated and stored in one table with a uniform schema.

**Right now, it’s clunky to:**

discover tokens via PoolCreated and then subscribe to their ERC-20 Transfer events, filtered only to rows where from/to == the discovered pool;

merge token0+token1 results into a single table without building custom plumbing.

**Feature request (two parts)**

_1. Discovered address filters_
Allow event subscriptions to reference addresses discovered upstream (factory outputs) and then apply field-level filters on downstream events.

_2. Union multiple subscriptions into one sink table_
Let multiple “child” subscriptions (e.g., token0 and token1 ERC-20 streams) write into one table with a shared schema, with built-in dedup on (tx_hash, log_index) and optional unique_keys override.

**Current implementation allows to catch only one input per factory and create separate table for each input captured. So we will have two identical transfers tables for `token0` and `token1` and we can't filter out Transfers by TX src/destination:**
```
- name: HyperswapV3TransfersToken0 
  details:
    - network: hyperevm
      start_block: 11648
      end_block: 100000
      factory:
        name: HyperswapV3Factory
        address: 0xb1c0fa0b789320044a6f623cfe5ebda9562602e3
        abi: ./abis/HyperswapV3Factory.abi.json
        event_name: PoolCreated
        input_name: "token0"
  abi: ./abis/ERC20.abi.json
  include_events:
  - Transfer
- name: HyperswapV3TransfersToken1
  details:
    - network: hyperevm
      start_block: 11648
      end_block: 100000
      factory:
        name: HyperswapV3Factory
        address: 0xb1c0fa0b789320044a6f623cfe5ebda9562602e3
        abi: ./abis/HyperswapV3Factory.abi.json
        event_name: PoolCreated
        input_name: "token1"
[...]

PR #287, opened 2025-08-17
Extend factory filtering to support array as `input_names`

desfero:
Fixes one part of #286

### DIFF
--- a/documentation/docs/pages/docs/start-building/yaml-config/contracts.mdx
+++ b/documentation/docs/pages/docs/start-building/yaml-config/contracts.mdx
@@ -198,7 +198,7 @@ The event name to filter on, it must match the ABI event name.
 
 #### input_name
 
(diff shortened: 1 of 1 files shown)


documentation/docs/pages/docs/start-building/yaml-config/contracts.mdx
```diff
@@ -198,7 +198,7 @@ The event name to filter on, it must match the ABI event name.
 
 #### input_name
 
-The path to the factory-deployed contract address in the event inputs. Supports deep property access in case of complex event types: `pool.address`.
+The path to the factory-deployed contract address in the event inputs. If a factory deploys more than one contract in a single event, you can provide an array of inputs. Supports deep property access in case of complex event types: `pool.address`.
```
joshstevens19: add an example of many vs single as not clear from just this

PR #293, opened 2025-08-27
Clickhouse DB Support

jaggad:


cli/src/commands/new.rs
```diff
@@ -153,11 +154,17 @@ pub fn handle_new_command(
     let repository = prompt_for_optional_input::<String>("Repository", None);
     let storage_choice = prompt_for_input_list(
         "What Storages To Enable? (graphql can only be supported if postgres is enabled)",
-        &["postgres".to_string(), "csv".to_string(), "both".to_string(), "none".to_string()],
+        &[
+            "postgres".to_string(),
+            "clickhouse".to_string(),
+            "csv".to_string(),
+            "postgres and csv".to_string(),
```
jaggad: im tempted to remove the combined `postgres and csv` as an option tbh

core/Cargo.toml
```diff
@@ -92,6 +92,8 @@ reth-exex = { git = "https://github.com/paradigmxyz/reth", tag = "v1.5.0", featu
 reth-node-api = { git = "https://github.com/paradigmxyz/reth", tag = "v1.5.0", optional = true }
 reth-node-ethereum = { git = "https://github.com/paradigmxyz/reth", tag = "v1.5.0", optional = true }
 reth-tracing = { git = "https://github.com/paradigmxyz/reth", tag = "v1.5.0", optional = true }
+async-std = "1.13.2"
+clickhouse = { version = "0.13.3", features = ["chrono"] }
```
jaggad: can probably drop the chrono dep, as we serialize via string using simple execute method rather than their native http protocol method

cli/src/commands/new.rs
```diff
@@ -279,6 +296,16 @@ pub fn handle_new_command(
     // Write the rindexer.yaml file
     write_manifest(&manifest, &rindexer_yaml_path)?;
 
+    // Write .env if required
+    if clickhouse_enabled {
+        let env = "CLICKHOUSE_URL=\nCLICKHOUSE_USER=\nCLICKHOUSE_PASSWORD=";
```
jaggad:
```suggestion
        let env = "CLICKHOUSE_URL=\nCLICKHOUSE_USER=\nCLICKHOUSE_PASSWORD=\nCLICKHOUSE_DB=";
```

core/src/database/clickhouse/client.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+impl ClickhouseClient {
+    pub async fn new() -> Result<Self, ClickhouseConnectionError> {
+        let connection = clickhouse_connection()?;
+
+        let client = Client::default()
+            .with_url(connection.url)
+            .with_user(connection.user)
+            .with_database(connection.db)
```
jaggad, line 54: Maybe make optional and unwrap to default
```
            .with_database(connection.db)
```

cli/src/commands/new.rs
joshstevens19: yeah tbh i dont see both needing to be used so feel free to remove

core/src/database/clickhouse/client.rs
```diff
@@ -0,0 +1,115 @@
...
+            sql.split(';').map(str::trim).filter(|s| !s.is_empty()).collect();
+
+        for statement in statements {
+            self.execute(statement).await?;
+        }
+
+        Ok(())
+    }
+
+    pub async fn bulk_insert(
```
joshstevens19: it may be worth a quick trait here with basic required methods at least then we start being able to understand bare bones of a db layer

jaggad: Think I will leave this for now since it seems the implementation for both clients will keep evolving for a bit. Probably makes sense at some point but for now they are different `impls` and technically could have different interfaces if we want to.

.gitignore
```diff
@@ -18,9 +18,11 @@ cli/rindexer.yaml
 cli/.env
 examples/*
 !examples/rindexer_demo_cli
+!examples/rindexer_clickhouse
```
desfero: what's the difference between `rindexer_clickhouse` and `rust_clickhouse`?

core/src/database/clickhouse/client.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    pub async fn query_one<T>(&self, sql: &str) -> Result<T, ClickhouseError>
+    where
+        T: Row + for<'b> Deserialize<'b>,
+    {
+        self.conn
+            .query(sql)
+            .fetch_one()
+            .await
+            .map_err(|e| ClickhouseError::ClickhouseError(e.to_string()))
```
desfero: question: is the error type always different between methods (`query_one`, `execute`, etc) that we do `e.to_string()` on the error?

core/src/database/clickhouse/client.rs
```diff
@@ -0,0 +1,124 @@
...
+    #[error("The clickhouse env vars are wrong please check your environment: {0}")]
+    ClickhouseConnectionConfigWrong(#[from] env::VarError),
+
+    #[error("Could not connect to clickhouse database: {0}")]
+    ClickhouseNetworkError(#[from] clickhouse::error::Error),
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum ClickhouseError {
+    #[error("ClickhouseError {0}")]
```
desfero:
```suggestion
    #[error("ClickhouseError: {0}")]
```

core/src/database/clickhouse/client.rs
```diff
@@ -0,0 +1,124 @@
...
+            .await
+            .map_err(|e| ClickhouseError::ClickhouseError(e.to_string()))
+    }
+
+    pub async fn execute_batch(&self, sql: &str) -> Result<(), ClickhouseError> {
+        let statements: Vec<&str> =
+            sql.split(';').map(str::trim).filter(|s| !s.is_empty()).collect();
+
+        for statement in statements {
+            self.execute(statement).await?;
```
desfero, line 92: question: how does it work under the hood, will it block next statements execution or it's more like fire and forget?

core/src/database/clickhouse/client.rs
```diff
@@ -0,0 +1,124 @@
...
+        Ok(())
+    }
+
+    pub(crate) async fn bulk_insert_via_query(
+        &self,
+        table_name: &str,
+        column_names: &[String],
+        bulk_data: &[Vec<EthereumSqlTypeWrapper>],
+    ) -> Result<u64, ClickhouseError> {
+        let column_names_str = column_names.join(", ");
```
desfero: very nit to avoid mutable vector
```suggestion
let values = bulk_data
    .iter()
    .map(|row| {
        let row_values = row.iter()
            .map(|v| v.to_clickhouse_value())
            .collect::<Vec<_>>()
            .join(", ");
        format!("({})", row_values)
    })
    .collect::<Vec<_>>()
    .join(", ");

let query = format!(
    "INSERT INTO {} ({}) VALUES {}",
    table_name,
    column_names.join(", "),
    values
);

self.execute(&query).await?;
Ok(bulk_data.len() as u64)```

File: core/src/database/clickhouse/setup.rs [hunk 11]
```diff
@@ -0,0 +1,61 @@
...
+        info!("Dropped all data for {}", manifest.name);
+    }
+
+    if disable_event_tables {
+        info!("Creating internal rindexer tables for {}", manifest.name);
+    } else {
+        info!("Creating tables for {}", manifest.name);
+    }
+
+    let sql = generate_tables_for_indexer_clickhouse(project_path, &manifest.to_indexer(), false)
```
desfero: `disable_event_tables` is not used except adding logs should we pass it to the `generate_tables_for_indexer_clickhouse`?

core/src/database/generate.rs
```diff
@@ -0,0 +1,359 @@
...
+            let table_name = format!("{}.{}", schema_name, camel_to_snake(&event_info.name));
+            info!("Creating table if not exists: {}", table_name);
+            let event_columns = if event_info.inputs.is_empty() {
+                "".to_string()
+            } else {
+                generate_columns_with_data_types(&event_info.inputs).join(", ") + ","
+            };
+
+            let create_table_sql = format!(
+                "CREATE TABLE IF NOT EXISTS {table_name} (\
```
desfero, line 35: formatting is off

core/src/database/clickhouse/generate.rs
```diff
@@ -0,0 +1,278 @@
...
+                input_names: factory.input_names(),
+            };
+            let table_name = generate_internal_factory_event_table_name(&params);
+
+            let create_table_query = format!(
+                r#"
+            CREATE TABLE IF NOT EXISTS rindexer_internal.{table_name} (
+                "factory_address" FixedString(42),
+                "factory_deployed_address" FixedString(42),
+                "network" TEXT
```
desfero: is there a `TEXT` type in clickhouse?

core/src/database/sql_type_wrapper.rs
desfero, line 1: we should move this big enum outside of `postgres` folder given it's now both postgres and clickhouse, maybe we keep it at the root of `database` folder and have just functions `get_clickhouse_type(EthereumSqlTypeWrapper)/get_postgresql_type(EthereumSqlTypeWrapper)` for postgresql/clickhouse inside respective folders

core/src/event/config.rs
```diff
@@ -38,6 +39,7 @@ pub struct ContractEventProcessingConfig {
     pub registry: Arc<EventCallbackRegistry>,
     pub progress: Arc<Mutex<IndexingEventsProgressState>>,
     pub database: Option<Arc<PostgresClient>>,
```
desfero: nit: should be now named `postgress` given both `postgress` and `clickhouse` is a database

core/src/generator/build.rs
```diff
@@ -250,7 +241,19 @@ pub fn generate_rindexer_typings(
...
+                write_file(
+                    &generate_file_location(&output, "database"),
+                    generate_postgres_code().as_str(),
+                )
+                .map_err(WriteGlobalError::from)?;
+            }
+
+            if manifest.storage.clickhouse_enabled() {
+                write_file(
+                    &generate_file_location(&output, "database"),
```
desfero, line 253: what if both are enabled, would write to the same file 🤔 ?

desfero: ok i see, not supported right now

core/src/generator/events_bindings.rs
```diff
@@ -972,6 +988,13 @@ pub fn generate_event_handlers(
                         return Err(e.to_string());
                     }}
                 "#,
+                // table_name = if storage.postgres_enabled() {
```
desfero: commented code 🤔 ?

core/src/indexer/no_code.rs
```diff
@@ -439,25 +462,45 @@ fn no_code_callback(params: Arc<NoCodeCallbackParams>) -> EventCallbacks {
...
-                        )
-                        .await
-                    {
-                        error!(
-                            "{}::{} - Error performing bulk insert: {}",
-                            params.contract_name, params.event_info.name, e
-                        );
-                        return Err(e.to_string());
+                    // anything over 100 events is considered bulk and goes the COPY route
+                    if bulk_data_length > 100 {
```
desfero: @joshstevens19 @jaggad this feels like a regression, why we don't write to the postgres if we get less than 100 events?

core/src/indexer/no_code.rs
```diff
@@ -439,25 +462,45 @@ fn no_code_callback(params: Arc<NoCodeCallbackParams>) -> EventCallbacks {
...
+                        if let Err(e) = postgres
+                            .insert_bulk(
+                                &params.sql_event_table_name,
+                                &params.sql_column_names,
+                                &sql_bulk_data,
+                            )
+                            .await
+                        {
+                            error!(
+                                "{}::{} - Error performing bulk insert: {}",
```
desfero:
```suggestion
                                "{}::{} - Error performing postgres bulk insert: {}",
```

core/src/indexer/no_code.rs
```diff
@@ -623,6 +667,7 @@ pub async fn process_events(
             project_path,
             manifest,
             postgres.clone(),
+            clickhouse.clone(),
```
desfero, line 667: we would need at some point to clean it up as given we support just single database at a time it would be nicer to use `enum` here as you would never end up in a situation where both `postgres` and `clickhouse` are `Some` therefore 
```
enum Database {
  ..
}

database: Option<Database>;
```
could better convey that message

core/src/manifest/stream.rs
```diff
@@ -62,9 +62,26 @@ pub struct RedisStreamStreamConfig {
     pub events: Vec<StreamEvent>,
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct ExchangeKindWrapper(pub ExchangeKind);
 
+impl Serialize for ExchangeKindWrapper {
```
desfero, line 68: out of curiosity why we decided to use custom implementation?

core/src/manifest/storage.rs
```diff
@@ -109,6 +120,9 @@ pub struct Storage {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub postgres: Option<PostgresDetails>,
 
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub clickhouse: Option<ClickhouseDetails>,
```
desfero, line 125: 🤔 what happens when someone tries to setup `postgres` and `clickhouse` at the same time?

documentation/docs/pages/docs/changelog.mdx
```diff
@@ -128,6 +128,7 @@ github branch - https://github.com/joshstevens19/rindexer/tree/release/0.25.0
 - feat: Adds nocode postgres migration "versioning system"
 - feat: Expose a rust project handler to process raw blocks with transactions if native indexing is enabled.
 - feat: bump alloy to 1.0.27
+- feat: Add clickhouse option to rindexer
```
desfero: should it be in `## Changes Not Deployed` section?

desfero:
```suggestion
- feat: Add clickhouse integration to rindexer
```

documentation/docs/pages/docs/start-building/yaml-config/storage.mdx
```diff
@@ -986,6 +986,157 @@ storage:
...
+storage:
+  clickhouse:
+    enabled: true
+    drop_each_run: true // [!code focus]
+```
+
+### disable_create_tables
+
+:::info
+This is only relevant for the rust projects as no-code if postgres is enabled will have to create the tables.
```
desfero:
```suggestion
This is only relevant for the rust projects as no-code if clickhouse is enabled will have to create the tables.
```

documentation/docs/pages/docs/start-building/yaml-config/storage.mdx
```diff
@@ -986,6 +986,157 @@ storage:
 
 You can read more about how this changes the GraphQL ability to query the data [here](/docs/accessing-data/graphql#relationships).
 
+
+## clickhouse
+
+If you wish to store the data in a clickhouse database with the no-code project you can enable the clickhouse storage.
```
desfero, line 992: Can you mix and match storage like with `postgres and csv`?

.gitignore
jaggad: `rindexer_clickhouse` is nocode, maybe i will rename to `nocode_clickhouse` since I realised prefixing everything with rindexer may not be required.

core/src/database/clickhouse/client.rs
jaggad: You're right, this is an oversight, fixing now.

core/src/database/clickhouse/client.rs
jaggad: This blocks each statement right now, as far as i know we don't have native clickhouse client support for simple-batch like postgres.

So this takes those "simple batch pipelines" and just runs them as consecutive queries.

It is only used in rindexer startup to create tables / do nothing, so it's very fast atm

core/src/database/clickhouse/setup.rs
jaggad: Yeah good catch, I think you're right here, it would always create the tables right now.

core/src/database/clickhouse/generate.rs
jaggad: Yeah there technically is, but it's just an alias of `String` so I'd rather use that. Good catch.

core/src/generator/events_bindings.rs
jaggad: can be deleted, leftover from testing yesterday

core/src/manifest/stream.rs
jaggad: Lapin doesn't implement Serde or have a feature flag for serde, so there's no derive possible on `ExchangeKindWrapper` and we cant deserialize/serialize `ExchangeKind`

documentation/docs/pages/docs/changelog.mdx
jaggad: Good point, must have auto-rebased here

documentation/docs/pages/docs/start-building/yaml-config/storage.mdx
jaggad: Negative, one storage enabled at a time. Also removed the `postgres + csv` option after discussing with Josh.

core/src/manifest/storage.rs
jaggad: I've now set it up to error on parsing the manifest so people have more instant feedback and we can be sure the invariant is applied.

added a unit test for this

core/src/generator/build.rs
jaggad: Enforced this invariant at the manifest parsing time now too, just to be sure.

core/src/indexer/no_code.rs
jaggad: I agree with the sentiment, main reason I wanted to avoid that for now is I think this PR sets the baseline for a proper `data-storage` layer refactor throughout rindexer.

Maybe bundled with separating things out into crates and allowing us to actually use a unified trait impl, or similar.

For now using two fields was the least impactful change so we could focus on core logic this time around.

core/src/indexer/no_code.rs
jaggad: This looks like a bad merge from yesterday, good find. I will address now.

core/src/event/config.rs
jaggad: Left as it was to impact the code less, but it probably does make sense to update, so have updated

core/src/database/sql_type_wrapper.rs
jaggad: Agree, for now I think we can just have it all in root.

PR #301, opened 2025-09-01
breaking: update `contract_address` column type and `insert_bulk` API

zannis:
Changes:

Made Postgres methods bulk_insert_via_copy and bulk_insert_via_query internal to the crate now, with insert_bulk being the public API for bulk inserts
Updated all playground indexer files to use insert_bulk instead of directly calling internal methods:
erc_20_filter.rs
playground_types_filter.rs
rocket_pool_eth.rs
uniswap_v3_pool_filter.rs
Updated no_code.rs to use insert_bulk instead of manual size-based logic
Code generators: Updated template generation in events_bindings.rs and trace_bindings.rs to generate code using insert_bulk
Cleaned up some error messages based on the changes above
Added sanity check so that insert_bulk Uses COPY method if rows > 100 OR total_params > 65535 (which is the Postgres limit for maximum params)
Updated contract_address generated column to be char(42)
Updated indexed_filters slice to contain 3 items
Updated generated examples to reflect latest code
Removed playground_types from rust_playground project as they were unused in rindexer.yaml


core/src/event/rindexer_event_filter.rs
```diff
@@ -180,25 +180,25 @@ impl RindexerEventFilter {
 
     pub fn topic1(&self) -> Topic {
         match self {
-            RindexerEventFilter::Address(filter) => filter.topics[1].clone(),
-            RindexerEventFilter::Filter(filter) => filter.topics[1].clone(),
-            RindexerEventFilter::Factory(filter) => filter.topics[1].clone(),
+            RindexerEventFilter::Address(filter) => filter.topics[0].clone(),
```
joshstevens19: was there a reason we changed this? was it breaking @zannis ?

zannis: discussed and reverted

joshstevens19 (approved):

Issue #313, opened 2025-09-25
[Feature Request] Cloudflare Queue streaming support


ind-igo:
Currently exploring the use of rindexer for realtime usage. I have most of my backend built on Cloudflare workers, and am in need of getting realtime updates from the chain propagated to my backend. To get realtime streaming now I would need to run a separate redis/kafka/rabbitmq instance, but being able to integrate directly into Cloudflare queues (and PubSub as well if possible) would be very helpful.

joshstevens19:
hey thanks for this so your hoping to do something like


```javascript
export default {
  async queue(batch, env, ctx) {
    // Process each message in the batch
    for (const message of batch.messages) {
      try {
        // Parse the blockchain event data
        const eventData = message.body;
        
        console.log('Processing blockchain event:', {
          eventName: eventData.event_name,
          network: eventData.network,
          messageId: eventData.message_id,
          eventCount: eventData.event_data?.length || 0
        });

        // Process Transfer events
        if (eventData.event_name === 'Transfer') {
          await processTransferEvents(eventData.event_data, env);
        }

        // Acknowledge the message
        message.ack();
      } catch (error) {
        console.error('Error processing message:', error);
        // Retry the message
        message.retry();
      }
    }
  }
};

async function processTransferEvents(transfers, env) {
  for (const transfer of transfers) {
    // Store high-value transfers in D1 database
    if (BigInt(transfer.value) > BigInt('1000000000000000000')) {
      await env.DB.prepare(
        'INSERT INTO transfers (from_address, to_address, value, block_number, tx_hash) VALUES (?, ?, ?, ?, ?)'
      ).bind(
        transfer.from,
        transfer.to,
        transfer.value,
        transfer.tx_information.block_number,
        transfer.tx_information.transaction_hash
      ).run();
    }
  }
}
```

with an config like

```yaml [contract events]
name: RocketPoolETHIndexer
description: My first rindexer project
repository: https://github.com/joshstevens19/rindexer
project_type: no-code
networks:
- name: ethereum
  chain_id: 1
  rpc: https://mainnet.gateway.tenderly.co
contracts:
- name: RocketPoolETH
  details:
  - network: ethereum
    address: "0xae78736cd615f374d3085123a210448e74fc6393"
    start_block: "18600000"
    end_block: "18600181"
  abi: "./abis/RocketTokenRETH.abi.json"
  include_events:
  - Transfer
  streams: // [!code focus]
    cloudflare_queues: // [!code focus]
[...]

PR #314, opened 2025-09-25
feat: add cloudflare queues to the streams

joshstevens19:


joshstevens19:
related - https://github.com/joshstevens19/rindexer/issues/313

PR #321, opened 2025-10-08
Do not log error if we are backwards, but in safe reorg range

desfero:
Cleanup provider handling
Throw error when a provider chain id is different from one in rindexer config

### DIFF
--- a/core/src/indexer/start.rs
+++ b/core/src/indexer/start.rs
@@ -155,8 +155,7 @@ async fn get_start_end_block(
     }
 
     let (end_block, indexing_distance_from_head) =
-        calculate_safe_block_number(reorg_safe_distance, &provider, latest_block, end_block)
-            .await?;
+        calculate_safe_block_number(reorg_safe_distance, &provider, latest_block, end_block);
 
     Ok((start_block, end_block, indexing_distance_from_head))
 }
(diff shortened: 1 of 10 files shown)


core/src/indexer/start.rs
```diff
@@ -630,22 +629,21 @@ pub async fn initialize_database(
...
     reorg_safe_distance: bool,
     provider: &Arc<JsonRpcCachedProvider>,
     latest_block: U64,
     mut end_block: U64,
-) -> Result<(U64, U64), StartIndexingError> {
+) -> (U64, U64) {
     let mut indexing_distance_from_head = U64::ZERO;
     if reorg_safe_distance {
-        let chain_id =
-            provider.get_chain_id().await.map_err(StartIndexingError::GetChainIdError)?;
```
desfero, line 642: was not optimal, there is no need to call `eth_chain_id` every safe block number calculation. 
Also this function is not a sync one with is another win

core/src/provider.rs
```diff
@@ -182,44 +180,6 @@ fn is_known_zk_evm_compatible_chain(chain: Chain) -> Option<bool> {
 }
 
 impl JsonRpcCachedProvider {
-    pub async fn new(
```
desfero, line 185: logic was spread across two separate functions
- `JsonRpcCachedProvider` method
- `create_provider` function

don't see a need for that as we can nicely abstract all in single function. `JsonRpcCachedProvider` has private fields that are only accessible in this module so there is also no other way to create the struct than by calling `create_provider` method

On top of that having `Struct::new()` is not fully idiomatic rust given constructors should be not async.

core/src/provider.rs
```diff
@@ -596,12 +556,6 @@ impl JsonRpcCachedProvider {
         Ok(filtered_logs)
     }
 
-    #[tracing::instrument(skip_all)]
-    pub async fn get_chain_id(&self) -> Result<U256, ProviderError> {
```
desfero, line 600: hiding this as should be never called by our code as affects negatively indexing speed

core/src/provider.rs
```diff
@@ -679,18 +635,43 @@ pub async fn create_client(
...
 
-    Ok(Arc::new(
-        JsonRpcCachedProvider::new(
-            provider,
+    let real_rpc_chain_id = provider.get_chain_id().await.map_err(|e| {
+        RetryClientError::CouldNotConnectClient(RpcError::LocalUsageError(Box::new(e)))
+    })?;
+
+    if real_rpc_chain_id != chain_id {
+        return Err(RetryClientError::InvalidClientChainId(
```
desfero, line 666: adding this check to make sure config `chain_id` matches the real RPC `chain_id`.
Tested this locally and if there is a mismatch indexer is going into a weird broken state.

PR #323, opened 2025-10-08
Add support for `RINDEXER_LOG` env variable

desfero:


core/src/event/config.rs
```diff
@@ -153,7 +156,7 @@ impl FactoryEventProcessingConfig {
     }
 
     pub fn info_log_name(&self) -> String {
-        format!("{}::{}", self.contract_name, self.event.name)
+        format!("{}::{}::{}", self.contract_name, self.event.name, self.network_contract.network)
```
desfero, line 159: adding network to `info_log_name` given `EventProcessingConfig` is already per network config

core/src/logger.rs
```diff
@@ -73,8 +73,12 @@ impl tracing_subscriber::fmt::time::FormatTime for CustomTimer {
     }
 }
 
-pub fn setup_logger(log_level: LevelFilter) {
-    let filter = EnvFilter::from_default_env().add_directive(log_level.into());
+const LOG_LEVEL_ENV: &str = "RINDEXER_LOG";
```
desfero, line 76: 👀

core/src/indexer/process.rs
```diff
@@ -42,13 +42,26 @@ pub enum ProcessEventError {
     ProviderCallError(#[from] ProviderError),
 }
 
-pub async fn process_event(
+/// Processes an event that doesn't have dependencies.
```
desfero, line 45: added extra description and renamed methods to be easier to understand, plus extra debugging logs that we need

core/src/event/config.rs
jaggad: Thoroughly approve of this change 👍

core/src/indexer/process.rs
```diff
@@ -516,12 +523,11 @@ async fn live_indexing_for_contract_event_dependencies(
                             let complete = task.await;
                             if let Err(e) = complete {
                                 error!(
-                                                        "{}::{} - {} - Error indexing task: {} - will try again in 200ms",
-                                                        &config.info_log_name(),
-                                                        &config.network_contract().network,
-                                                        IndexingEventProgressStatus::Live.log(),
-                                                        e
-                                                    );
+                                    "{} - {} - Error indexing task: {} - will try again in 200ms",
```
jaggad, line 526: I'm inclined to say anything that self-recovers could be a `warn!`, whereas messages like `RPC has gone back on latest block` need to be an error since it is unliekly to self-recover

PR #331, opened 2025-10-17
Adding hosted clickhouse support + changing naming convention for cli…

kakagri:
fixing issue with hosted clickhouse databases + naming issue, this is 100% vibe coded so be careful when merging :)

### DIFF
--- a/core/src/database/clickhouse/generate.rs
+++ b/core/src/database/clickhouse/generate.rs
@@ -36,7 +36,11 @@ pub fn generate_tables_for_indexer_clickhouse(
         if !disable_event_tables {
             info!("Creating event tables in database: {}", database_name);
 
-            sql.push_str(&generate_event_table_clickhouse(&event_names, database_name, &schema_name));
+            sql.push_str(&generate_event_table_clickhouse(
+                &event_names,
+                database_name,
+                &schema_name,
+            ));
         }
 
         sql.push_str(&generate_internal_event_table_clickhouse(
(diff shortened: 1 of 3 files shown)


core/src/database/clickhouse/generate.rs
```diff
@@ -186,24 +194,30 @@ pub fn generate_columns_with_data_types(inputs: &[ABIInput]) -> Vec<String> {
     generate_columns(inputs, &GenerateAbiPropertiesType::ClickhouseWithDataTypes)
 }
 
-pub fn drop_tables_for_indexer_clickhouse(project_path: &Path, indexer: &Indexer) -> Code {
+pub fn drop_tables_for_indexer_clickhouse(project_path: &Path, indexer: &Indexer, database_name: &str) -> Code {
     let mut sql = String::new();
 
-    sql.push_str("DROP TABLE IF EXISTS rindexer_internal.latest_block;");
+    sql.push_str(&format!("DROP TABLE IF EXISTS {}.rindexer_internal_latest_block;", database_name));
```
jaggad: To clarify this is because in hosted environemnt you want to just have `default` as the database and have tables be on the main db right

cli/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rindexer_cli"
-version = "0.28.0"
+version = "0.28.1"
```
jaggad, line 3: Not sure we should bump this.

documentation/docs/pages/docs/changelog.mdx
```diff
@@ -17,6 +17,13 @@
 
 all release branches are deployed through `release/VERSION_NUMBER` branches
 
+# 0.28.1-beta - 17th October 2025
+
+### Bug fixes
+-------------------------------------------------
+- fix: add tls for hosted clickhouse databases 
+- fix: ClickHouse now uses single database from `CLICKHOUSE_DB` env var instead of requiring `CREATE DATABASE` privileges, rationale: giving all permissions to one script to drop any database is unreasonable.
```
jaggad, line 25: Add these changes under the `## Changes Not Deployed` section rather than making a new release.

This allows multiple pending changes to be deployed with a new version

jaggad, line 25: > ClickHouse now uses single database from `CLICKHOUSE_DB` env var instead of requiring `CREATE DATABASE` privileges, rationale: giving all permissions to one script to drop any database is unreasonable.

Rationale makes a lot of sense to me, is a reasonable DBA perspective for permissions management. But this can be achieved by granting more granular permissions.

core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "rindexer"
-version = "0.28.0"
+version = "0.28.1"
```
jaggad, line 3: Don't bump here
```suggestion
version = "0.28.0"
```

core/src/indexer/last_synced.rs
```diff
@@ -183,11 +183,12 @@ pub async fn get_last_synced_block_number(config: SyncConfig<'_>) -> Option<U64>
             last_synced_block: u64,
         }
 
+        let database_name = clickhouse.get_database_name();
```
jaggad, line 186: Can we just call `clickhouse.database_name` instead of the getter fn

jaggad (changes requested):

PR #352, opened 2026-01-05
Expose a stream with indexer state changes

desfero:


core/src/manifest/core.rs
```diff
@@ -200,11 +200,7 @@ impl Manifest {
     }
 
     pub fn has_any_contracts_live_indexing(&self) -> bool {
-        self.all_contracts()
-            .iter()
-            .filter(|c| c.details.iter().any(|p| p.end_block.is_none()))
-            .count()
-            > 0
+        self.all_contracts().iter().any(|c| c.details.iter().any(|p| p.end_block.is_none()))
```
desfero, line 203: just nit, we don't need to loop through all values in the array

core/src/start.rs
```diff
@@ -269,30 +271,27 @@ pub async fn start_rindexer(details: StartDetails<'_>) -> Result<(), StartRindex
...
-                            &dependencies,
-                            false,
-                            indexing_details
-                                .registry
-                                .reapply_after_historic(processed_network_contracts),
-                            indexing_details.trace_registry.complete(),
-                        )
-                        .await
-                        .map_err(StartRindexerError::CouldNotStartIndexing)?;
+                if manifest.has_any_contracts_live_indexing() {
```
desfero, line 276: This fixes the issue that was affecting historical indexing when there was no relationship 
```
   // we index all the historic data first before then applying FKs
   !relationships.is_empty(),
```
Now historical indexing duration log is always there which makes codebase more cohesive

joshstevens19 (approved):

PR #368, opened 2026-02-03
fix: use calculated end_block for native_transfer_block_fetch

### LINKED ISSUE #367: native_transfer_block_fetch loops forever when end_block not set in manifest, blocking live indexing
When `native_transfers` is enabled without an `end_block` in the manifest, [start_historical_indexing()](cci:1://file:///home/outerlook/random-codes/joshstevens19/rindexer/core/src/indexer/start.rs:535:0-560:1) never completes because [native_transfer_block_fetch()](cci:1://file:///home/outerlook/random-codes/joshstevens19/rindexer/core/src/indexer/native_transfer.rs:99:0-169:1) receives `network_details.end_block = None` and loops forever.

**Root Cause:**
`@core/src/indexer/start.rs:266` passes `network_details.end_block` (raw manifest value) instead of calculated `end_block` to [native_transfer_block_fetch()](cci:1://file:///home/outerlook/random-codes/joshstevens19/rindexer/core/src/indexer/native_transfer.rs:99:0-169:1).

**Impact:**
- Historical indexing never completes
- [start_live_indexing()](cci:1://file:///home/outerlook/random-codes/joshstevens19/rindexer/core/src/indexer/start.rs:562:0-572:1) is never called
- All contract events (including swaps) are never indexed
- Native transfers loop forever

**Fix:**
Change line 266 from:
```rust
network_details.end_block,
```
to:
```rust
Some(end_block),
```

This aligns native transfer behavior with contract events, which respect `no_live_indexing_forced` and stop at the calculated `end_block` during historical phase.

outerlook:
## Summary

Pass the calculated `end_block` from `get_start_end_block()` instead of `network_details.end_block` to `native_transfer_block_fetch()`.

## Problem

When `end_block` is not set in the manifest, `network_details.end_block` is `None`, causing `native_transfer_block_fetch` to loop forever waiting for new blocks. This blocks `try_join_all` in `start_indexing`, preventing:
- Historical indexing from completing
- Live indexing from ever starting
- All contract events from being indexed

## Fix

Use the calculated `end_block` (which defaults to `latest_block` when not specified in manifest) instead of the raw manifest value. This aligns native transfer behavior with contract events.

Fixes #367

### DIFF
--- a/core/src/indexer/native_transfer.rs
+++ b/core/src/indexer/native_transfer.rs
@@ -191,15 +191,18 @@ pub async fn native_transfer_block_processor(
         }
 
         // Fetch more only if buffer was processed ok last time and cleared.
-        let recv = if buffer.is_empty() {
+        let was_empty = buffer.is_empty();
+        let recv = if was_empty {
             block_rx.recv_many(&mut buffer, concurrent_requests).await
         } else {
             buffer.len()
         };
 
-        if recv == 0 {
-            sleep(Duration::from_secs(1)).await;
-            continue;
+        // When recv_many returns 0 and buffer was empty, it means the channel is closed
+        // (all senders have been dropped). Exit the loop to complete historical indexing.
+        if recv == 0 && was_empty {
+            info!("Native transfer block processor completed - channel closed");
+            break Ok(());
         }
 
         let processed_block = native_transfer_block_consumer(
--- a/core/src/indexer/start.rs
+++ b/core/src/indexer/start.rs
@@ -263,7 +263,7 @@ async fn start_indexing_traces(
             network_details.cached_provider.clone(),
             block_tx,
             start_block,
-            network_details.end_block,
+            Some(end_block),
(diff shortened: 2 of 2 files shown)


jaggad:
@outerlook 

> When end_block is not set in the manifest, network_details.end_block is None, causing native_transfer_block_fetch to loop forever waiting for new blocks. This blocks try_join_all in start_indexing, preventing:

Can you please confirm by sharing logs or that it was validated in testing that the following conditions hold true with the current changes:

**When start block _is_ defined:**
1. `start_block: N` defined for all contracts and native transfer indexing (as such historical indexing will be needed here)
2. it completes historical indexing up to current head block
3. Then. Does it still continue indexing native transfers "live" in this setup? <--- this point to be confirmed

**When start block is not defined:**
1. No start block defined
2. Historical indexing process should skip or finish quickly
4. Live indexing does in fact index native transfers <--- this point to be confirmed

PR #371, opened 2026-02-07
feat: add support for tuple[] (array of structs) via JSONB storage

PickleyD:
This PR adds support for indexing events that contain tuple[] parameters (arrays of structs). Previously, these would cause a panic with the error "CustomStruct and Function are not supported yet".

### Problem
Events with `tuple[]` parameters cannot be flattened into separate database columns like single tuples because the array length varies per event. For example:

```
event ChainAddedOrSet(
    string indexed caip2ChainId,
    string assetRecoveryAddress,
    Account[] accounts  // tuple[] - array length varies
);

struct Account {
    string accountAddress;
    uint8 childContractScope;
}
```

### Solution
Store `tuple[]` as a JSONB column in PostgreSQL. This:

- Preserves all data without changing rindexer's one-row-per-event model
- Is queryable via PostgreSQL's JSONB operators
- Requires no schema changes for varying array lengths

Example output in the database:

`[{"accountAddress": "0x1234...", "childContractScope": "0"}, {"accountAddress": "0x5678...", "childContractScope": "1"}]`

### Changes
`core/src/database/sql_type_wrapper.rs`:

- Added `tuple_to_json_value()` helper function to serialize tuple/struct values to JSON
- Extended array handling to serialize `tuple[]` as JSONB instead of panicking
- Handles both `DynSolValue::Tuple` and `DynSolValue::CustomStruct` variants (the latter appears when `eip712` feature is transitively enabled, e.g., via the `reth` feature)

`core/src/database/postgres/generate.rs`:

- Added `tuple[]` → `JSONB` type mapping

`core/src/abi.rs`:

- Skip flattening for `tuple[]` types - keep as single JSONB column

### Why handle both Tuple and CustomStruct?
When alloy's `eip712` feature is enabled (transitively via `reth` → `reth-rpc-eth-api`), arrays of structs are decoded as `DynSolValue::CustomStruct` rather than `DynSolValue::Tuple`. Both contain the same underlying data, just with different representations. The fix handles both to ensure it works regardless of which features are enabled.

### Testing
Added 6 unit tests covering:

- `tuple_to_json_value` with various field types (string, address, bool, uint)
- Array of `DynSolValue::Tuple` serialization
- Array of `DynSolValue::CustomStruct` serialization
- Empty array handling

Run tests with:

`cargo test --package rindexer --lib --features "jemalloc,reth" -- database::sql_type_wrapper::tests`
Note: Tests require `--features "jemalloc,reth"` because `CustomStruct` is only available when `eip712` is enabled.


core/src/database/sql_type_wrapper.rs
```diff
@@ -1950,3 +2022,200 @@ pub fn map_ethereum_wrapper_to_json(
...
+                let second = arr[1].as_object().unwrap();
+                assert_eq!(second.get("accountAddress").unwrap(), "0xBBBB");
+                assert_eq!(second.get("childContractScope").unwrap(), "2");
+            }
+            _ => panic!("Expected JSONB wrapper"),
+        }
+    }
+
+    #[test]
+    fn test_map_log_token_for_empty_tuple_array() {
```
ExoMonk, line 2220: I think while empty tuple[], it produces wrong wrapper type (column count mismatch)
-> When the array is empty, `tokens.first()` is `None`,so exec enters the fallback branch which calls `tuple_solidity_type_to_ethereum_sql_type_wrapper(components)`. 

so we have flattened scalar wrappers  like `[String(""), U8(0)]`) instead of a single `JSONB(Array([]))`. The DB column expects one JSONB value but receives N scalars and could produce tunitime panic or data being corruptd

this test only asserts `!result.is_empty()` and doesn't verify the result is `JSONB(Array([])),` which I think masks the issue

core/src/database/sql_type_wrapper.rs
```diff
@@ -1323,6 +1323,60 @@ fn process_tuple(abi_inputs: &[ABIInput], tokens: &[DynSolValue]) -> Vec<Ethereu
...
+                            Value::String(format!("0x{}", hex::encode(b)))
+                        }
+                        DynSolValue::Tuple(nested_tuple) => {
+                            if let Some(nested_components) = &component.components {
+                                tuple_to_json_value(nested_components, nested_tuple)
+                            } else {
+                                Value::Null
+                            }
+                        }
+                        _ => Value::Null,
```
ExoMonk, line 1376: the inner `Array` match handles `DynSolValue::Tuple` but not `DynSolValue::CustomStruct` 
falling here

IMO when eip-712 is enabled and there's an array inside a tuple, struct elements become null ?
Isnt `CustomStruct` becomes silently null

core/src/abi.rs
```diff
@@ -116,21 +116,33 @@ impl ABIInput {
...
-                        },
-                    );
-
-                    ABIInput::generate_abi_name_properties(
-                        components,
-                        properties_type,
-                        Some(new_path),
-                    )
-                } else {
+                    if is_tuple_array {
```
ExoMonk:
```
if !is_tuple_array {
    return ABIInput::generate_abi_name_properties(...);
}
```

WDYT ? I thought it was too silent

core/src/database/postgres/generate.rs
```diff
@@ -80,6 +80,12 @@ pub fn solidity_type_to_db_type(abi_type: &str) -> String {
     let is_array = abi_type.ends_with("[]");
     let base_type = abi_type.trim_end_matches("[]");
 
+    // Handle tuple types as JSONB (both single tuples and arrays of tuples)
+    // Arrays of tuples cannot be flattened into columns since array length varies
+    if base_type == "tuple" && is_array {
```
ExoMonk: ClickHouse's `solidity_type_to_clickhouse_type` still has `_ => panic!()` for tuple

Since the PR changes token-level behavior to no longer panic early, the failure now occurs deeper in the pipeline, which is more confusing IMO
Just add a clear error msg I think

core/src/abi.rs
```diff
@@ -116,21 +116,33 @@ impl ABIInput {
         inputs
             .iter()
             .flat_map(|input| {
+                // For tuple[] (array of structs), don't flatten - store as JSONB
+                // Array length varies so we can't pre-determine column count
+                let is_tuple_array = input.type_ == "tuple[]";
```
ExoMonk: uses strict equality here, don't Solidity `tuple[3]` won't match and will still attempt the old flattening path, likely paniccing ?

`solidity_type_to_db_type` uses ends_with("[]") which also misses fixed-size arrays

joshstevens19 (approved):

PR #380, opened 2026-03-08
Parallel Range Fetching for Historical Backfill

ExoMonk:
## Summary
Atm, hist backfill in rindexer is sequential: 1 `eth_getLogs` [call at a time, wait for response, process, repeat](https://github.com/joshstevens19/rindexer/blob/master/core/src/indexer/fetch_logs.rs#L80). 
For large block ranges (millions of blocks), this means the indexer spends most of its time waiting on RPC round-trips. 
I thin k rindexer is leaving bandwidth on the table

I am considering adding parallel range fetching with write coalescing: configurable via a single `fetch_concurrency` field in config. The block range is split into chunks, and N workers fetch concurrently while a reorder buffer guarantees strict block ordering before delivery to the consumer.

Capture d’écran 2026-03-08 à 17 09 09

## Usage:

```
config:
  fetch_concurrency: 4
```

## Design & decision tTB discussed

- Reorder buffer (`BTreeMap`): Workers complete out-of-order, but events must be delivered in strict block order. The reorder buffer holds out-of-order results and flushes them in sequence.
-  `WorkerDropGuard` with `CancellationToken`: If a worker panics, the drop guard sends an error batch via `try_send()`. If the channel is full, it cancels the entire pipeline via `CancellationToken` to prevent silent data gaps.
- Notify-gated dispatcher: Uses `tokio::sync::Notify` instead of polling. Registers `notified()` future BEFORE checking condition to prevent lost-notification race.
- Adaptive concurrency (AIMD): Reuses the existing `AdaptiveConcurrency` controller per-pipeline. Workers respect backoff and scale down on rate limits.
- Write coalescing: Consumer uses `now_or_never()` (cancel-safe on tokio mpsc) to drain all ready results in a single batch before writing to storage.
- Graceful fallback: Ranges < 1000 blocks or factory contracts automatically use sequential path. No fetch_concurrency config = sequential (backwards compatible).

### DIFF
--- a/cli/src/commands/new.rs
+++ b/cli/src/commands/new.rs
@@ -219,6 +219,7 @@ pub fn handle_new_command(
             callback_concurrency: None,
             timestamp_sample_rate: None,
             max_concurrent_view_calls: None,
+            fetch_concurrency: None,
         },
         constants: HashMap::new(),
         timestamps: None,
--- a/core/src/indexer/fetch_logs.rs
+++ b/core/src/indexer/fetch_logs.rs
@@ -1,3 +1,4 @@
+use crate::adaptive_concurrency::AdaptiveConcurrency;
 use crate::blockclock::BlockClock;
 use crate::helpers::{halved_block_number, is_relevant_block};
 use crate::indexer::reorg::reorg_safe_distance_for_chain;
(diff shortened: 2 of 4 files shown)


joshstevens19:
replaced by https://github.com/joshstevens19/rindexer/pull/413

PR #381, opened 2026-03-12
feat: `BlockIndexingCompleted` event

desfero:


core/src/event/config.rs
```diff
@@ -37,7 +36,7 @@ pub struct ContractEventProcessingConfig {
     pub start_block: U64,
     pub end_block: U64,
     pub registry: Arc<EventCallbackRegistry>,
-    pub progress: Arc<Mutex<IndexingEventsProgressState>>,
+    pub progress: Arc<IndexingEventsProgressState>,
```
desfero, line 39: moving mutex to `IndexingEventsProgressState` to not leak internal details

PR #385, opened 2026-03-23
fix: use composite key in progress tracker for multi-network events

mgrabina:
## Problem

When the same contract is indexed on multiple networks (e.g. a `Spoke` contract on both `ethereum` chain 1 and a Tenderly fork), the `IndexingEventsProgressState` in `core/src/indexer/progress.rs` fails to track progress correctly.

### Root cause

In `monitor()`, the `events` HashMap uses `event_info.id` as the key. But when iterating over networks for each event, the same key is inserted twice:

```rust
for event_info in event_information {
    for network_contract in &event_info.contract.details {
        // Same event_info.id for both networks:
        events.insert(event_info.id.to_string(), /* chain-specific data */);
    }
}
```

The second network's entry **overwrites** the first. Later, when `update_last_synced_block(chain_id, id)` is called for the first network, the chain_id filter fails because the stored entry has the second network's chain_id:

```rust
let event = events.get_mut(id)
    .filter(|e| e.chain_id == chain_id)  // FAILS: stored chain_id is from the other network
    .ok_or_else(|| SyncError::EventNotFound(...))?;
```

This causes `EventNotFound` errors for every event on the overwritten chain, preventing `BlockIndexingCompleted` from ever firing for that chain.

### Impact

Any consumer of `BlockIndexingCompleted` for the affected chain never receives events:
- Asset history rebuild tasks never run
- Block sync timeouts on every user/reserve refresh (60s delays)
- Background tasks that depend on confirmed blocks are stuck

### Reproduction

Configure two networks for the same contract in `rindexer.yaml`:

```yaml
- name: Spoke
  details:
    - network: ethereum      # chain_id: 1
      address: ["0x..."]
      start_block: 24687800
    - network: fork           # chain_id: 123456789
      address: ["0x..."]
      start_block: 23697608
```

Run the indexer. After historical sync completes, the live indexing phase produces:
```
ERROR Error updating in-mem last synced block result: EventNotFound("1::PPu1BmHXwe")
ERROR Error updating in-mem last synced block result: EventNotFound("1::904yjM3oUI")
```

## Fix

Use a composite key `{chain_id}::{event_id}` instead of bare `event_id` for both the `events` map and the `block_networks[chain].events` map. This makes each (chain, event) pair unique.

The composite key is constructed via a `progress_key(chain_id, id)` helper, applied in:
- `monitor()` when registering events and trace events
- `update_last_synced_block()` when looking up events
- `update_event()` when mutating event state
- The min-block filter when computing `BlockIndexingCompleted`

The `block_networks` map continues to use the original event ID (not the composite key) since it's already keyed by chain_id.

### Test updates

- Existing tests updated to use composite keys in the test helper
- New test: `test_same_event_id_different_chains` verifies that the same event ID on two chains tracks progress independently and emits `BlockIndexingCompleted` for both

### DIFF
--- a/core/src/indexer/progress.rs
+++ b/core/src/indexer/progress.rs
@@ -287,7 +287,10 @@ impl IndexingEventsProgressState {
                     match events.get(key.as_str()) {
                         Some(e) => !matches!(e.status, IndexingEventProgressStatus::Completed),
                         None => {
-                            warn!("BlockProgress: event {} missing from events map for chain {}", id, report.chain_id);
+                            warn!(
+                                "BlockProgress: event {} missing from events map for chain {}",
+                                id, report.chain_id
+                            );
                             false
                         }
                     }


mgrabina:
@codex

PR #396, opened 2026-04-02
fix: handle int/uint array types in parse_solidity_integer_type

### LINKED ISSUE #318: Parsing int/uint arrays in events (codegen indexer)
Hello! sorry if this is wrong or misleading but i tried to use rindexer for a rust project and had errors while running 'rindexer codegen indexer'.
  To reproduce it you can start a new project, add contract 0xC59b0e4De5F1248C1140964E0fF287B192407E0C (mainnet) and try to run the above mentioned command.
The problems seems to be the helper function parse_solidity_integer_type which doesn't handle arrays for int & uint.
Wonder if im doing something wrong or this is a bug.
PD: no-code project runs correctly (but can't see the code generated)
PD2: rindexer codegen typings runs OK
Thanks for your time!

liu-zhipeng:
Strip array suffixes (e.g., [], [N]) before parsing the bit width to prevent ParseIntError when codegen encounters event definitions with integer array types like uint256[].

Fixes #318


ExoMonk (commented):
Few unit tests + E2E are welcome for this

liu-zhipeng:
Hi @ExoMonk  i pushed unit tests for my changes.

joshstevens19 (approved):

PR #401, opened 2026-04-10
feat: complete reorg handling

zannis:
## Reactive Reorg Handling for rindexer

rindexer is now reactive to chain reorganizations. Instead of relying solely on safe distance offsets to avoid reorged data, the indexer actively monitors the block hash chain and catches reorgs the moment they happen. When one is detected, indexing pauses, stale state is corrected, and indexing resumes — all within seconds.

The existing `reorg_safe_distance` configuration remains fully supported. Users who prefer the conservative approach of only indexing finalized blocks can continue using it as before — the two mechanisms are complementary. Safe distance prevents indexing unconfirmed blocks; reorg handling corrects state if a reorg slips through. They can be used independently or together.

This means three operating modes for downstream consumers:

**For latency-sensitive streams** (webhooks, Kafka, SNS, etc.) — events are delivered instantly as before, but now with a safety net. If a reorg invalidates previously-delivered events, a reorg notification is published through the same channels so consumers can reconcile. This is the `instant` delivery mode (default).

**For correctness-critical streams** — a new `finalized` delivery mode buffers events until they're past the chain's finality window before publishing. Events that survive the reorg safe distance are guaranteed canonical. No reconciliation needed.

**For maximum safety** — combine `reorg_safe_distance` on the contract with `reorg_handling` on the network. The safe distance delays indexing by N blocks, and reorg handling catches anything that slips past that window. Belt and suspenders.

```yaml
networks:
  - name: ethereum
    rpc: https://...
    reorg_handling:
      enabled: true       # active detection + correction
      window_size: 256

contracts:
  - name: MyContract
    reorg_safe_distance: true  # also wait N blocks before indexing (still works as before)
    # ...

streams:
  - type: webhook
    endpoint: https://example.com/realtime
    delivery: instant      # fast, with reorg notifications

  - type: kafka
    brokers: [...]
    delivery: finalized    # delayed, but guaranteed correct
```

### How it works

A per-network `ReorgCoordinator` validates the parent hash of every new block against a persisted sliding window of recent block hashes. On mismatch:

1. Fork point is identified via a single batch RPC call against the window
2. Stale events are deleted and affected tx hashes collected (single atomic PG transaction)
3. Block hash window and indexing checkpoints are corrected
4. `on_reorg` callback fires with invalidated tx hashes for user-side reconciliation
5. Reorg notification published through instant-mode streams
6. Indexing filter rewinds to the fork point — the next loop iteration re-fetches corrected events naturally

The entire recovery is blocking — no callbacks or dependent events see stale data.

Three detection paths converge on the same recovery flow:
- **Parent hash validation** (RPC polling) — catches reorgs on every new block
- **Removed logs** (`log.removed == true` from RPC) — catches reorgs surfaced during log fetching
- **Reth ExEx notifications** — instant detection when running an embedded reth node

On restart, startup validation compares the persisted window against the canonical chain and handles any reorgs that occurred while the indexer was offline.

### Aggregation table rollback

Reorg recovery extends to no-code aggregation/derived tables (`Set`, `Add`, `Increment`, `Subtract`, `Max`, `Min`). Two reversal strategies ensure derived state is always consistent:

- **Reversible operations** (`Add`/`Subtract`/`Increment`/`Decrement`) — a pre-deletion snapshot captures the aggregated values in the reorged block range, then inverse arithmetic is applied to the derived table.
- **Non-reversible operations** (`Set`/`Max`/`Min`) — an operation journal replays all surviving writes to recalculate the correct value from scratch.

### Native transfer rollback

Native transfer events (`EvmTraces.NativeTransfer`) are automatically registered in the reorg coordinator and participate in the same rollback/re-index flow as user-defined contract events.

### SQL injection defenses

All user-supplied identifiers from YAML specs are validated at startup before any SQL interpolation:

- `validate_sql_identifier` — table/column names must match `[a-zA-Z_][a-zA-Z0-9_]*`
- `validate_sql_value` — string literals (network names) restricted to alphanumeric, underscore, hyphen
- `validate_sql_condition` — filter expressions reject injection patterns (`--`, `/*`, `;`, quotes) and dangerous keywords (`UNION`, `DROP`, `DELETE`, etc.)

### What changed architecturally

The monolithic `reorg.rs` (1,790 lines) is replaced by a modular `reorg/` module:
- `coordinator.rs` — Per-network reorg detection and orchestration
- `task.rs` — Rollback execution (delete, correct, rewind, notify)
- `window.rs` — In-memory `BlockChainWindow` backed by a persisted `latest_blocks` table
- `persistence.rs` — DB read/write for the block hash window

A `ReorgContext` struct bundles DB clients, callback registry, and stream clients through the detection → recovery chain.

Reorg coordinators are only instantiated during live indexing — historical backfill skips them entirely, reducing memory overhead and correctly separating concerns.

### Crash recovery

**PostgreSQL** — the entire rollback (event deletion, block hash correction, checkpoint rewind) executes in a single atomic transaction. If the process crashes mid-recovery, the transaction rolls back and startup validation re-detects the reorg on next start.

**ClickHouse** — ClickHouse lacks multi-statement transactions, so the rollback is ordered for crash safety: checkpoint rewind executes first (ensuring re-indexing on restart), followed by event deletion via synchronous mutations (`mutations_sync = 1`), then block hash correction. Each step is idempotent, so a crash mid-sequence is retried from the start on next boot.

### Error handling

All reorg modules use `anyhow::Result` with `.with_context()` for structured error propagation — no string errors. Arithmetic uses `saturating_sub` where underflow is possible. `u64` → `i64` conversions for PostgreSQL use `try_from` with explicit error context.

### Multiple on_reorg callbacks

The callback registry supports multiple independent `on_reorg` handlers. `ReorgNotification` is `Clone`, enabling broadcast to all registered callbacks for flexible monitoring/alerting chains.

### New metrics

- `rindexer_reorg_handling_duration_seconds{network}`
- `rindexer_reorg_events_deleted_total{network}`
- `rindexer_reorg_detection_source_total{network, source}`
- `rindexer_reorg_cascade_total{network}`

### Test coverage

- 40 unit tests covering window operations, persistence round-trips, SQL validation, and derived table reversal strategies
- 29 integration tests using testcontainers (PostgreSQL, ClickHouse, Anvil) covering single-block reorg, deep reorg, multi-table, cross-chain, consecutive reorgs, checkpoint rewind, aggregation table rollback, journal recalculation, and offline recovery
- E2E tests for webhook stream delivery during reorg and reorg detection on startup after offline period


core/src/indexer/reorg/task.rs
```diff
@@ -0,0 +1,231 @@
...
+                    vec![]
+                }
+            }
+        } else {
+            vec![]
+        };
+
+        // Update the in-memory window with canonical blocks
+        if !canonical.is_empty() {
+            window.update_range(&canonical);
```
ExoMonk: called BEFORE ClickHouse rollback. I think that if CH fails after PG succeeds, the window marks the reorg as recovered but CH still has stale data

mauybe move it after both DB rollbacks succeed

core/src/indexer/fetch_logs.rs
```diff
@@ -553,64 +514,33 @@ async fn live_indexing_stream(
...
+                    if let Some(ref mut coordinator) = reorg_coordinator {
+                        let log_prefix = format!(
+                            "{} - {}",
+                            info_log_name,
+                            IndexingEventProgressStatus::live_log()
+                        );
+                        let reorg_ctx = ReorgContext {
+                            postgres: postgres.as_deref(),
+                            clickhouse: clickhouse.as_ref(),
+                            registry: None,
```
ExoMonk: Users who register `on_reorg` callbacks via code-gen will never receive notifications with registry set to None hardcode here ?
Maybe pass the `EventCallbackRegistry` item through to `ReorgContext`

core/src/indexer/reorg/task.rs
```diff
@@ -0,0 +1,231 @@
...
+                        self.fork_point,
+                        dt.full_table_name
+                    ),
+                    Err(e) => tracing::error!(
+                        "ClickHouse: failed to delete derived table rows in {}: {:?}",
+                        dt.full_table_name,
+                        e
+                    ),
+                }
+            }
```
ExoMonk, line 842: derived table rollback only deletes, no replay (accumulative upserts lose state) 
`set_derived_tables()` should be called & hanlded

core/src/indexer/reorg/task.rs
```diff
@@ -0,0 +1,231 @@
...
+        tracing::info!(
+            network = %self.network,
+            events_deleted = total_deleted,
+            duration_secs = duration,
+            "Reorg task completed"
+        );
+
+        Ok(ReorgTaskResult {
+            events_deleted: total_deleted,
+            events_reindexed: 0,
```
ExoMonk: dead pixel ?

zannis: good catch, fixing

core/src/indexer/reorg/task.rs
zannis: 👍🏼  moved after both rollbacks

core/src/indexer/fetch_logs.rs
zannis: fixed, exposed EventCallbackRegistry to ReorgContext

core/src/indexer/reorg/task.rs
zannis: I removed `events_reindexed` completely as events are actually reindexed later on

core/src/indexer/reorg/task.rs
zannis: Working on this one now - some cases are not fully possible to be handled automatically - im investigating adding a user-made reorg handling strategy on tables we cannot automatically revert

zannis: OK this is now done and all aggregation actions are now automatically managed, there are 2 cases:
* for reversible actions (Add, Subtract, Increment, Decrement) -> we apply the reverse action inside the reorg task
* for non reversible actions (Set, Min, Max) -> we store the state diffs on a new table for every event for the number of safe blocks per chain and once a reorg is found, we restore the state to the latest safe block, then apply all the state changes from the new healthy blocks one by one before continuing

zannis: updated the decription to reflect this

ExoMonk:
Sounds good, just E2E Tests failing here

ExoMonk (approved):

PR #402, opened 2026-04-10
feat: Add database field to custom tables: shared database override

ExoMonk:
New optional database field on no-code table definitions. When set, the custom table is created in the specified    
  database/schema instead of the default `{project}_{contract}` database. Raw event tables are unaffected.

```
  tables:                                                                                                             
    - name: events
      database: indexer  # ← all contracts write to indexer.events                                                    
      columns: ...                                          
````
                                                                                                             
Without database:, behavior is unchanged: tables go to `{project}_{contract}.{table}` as before.                     
                                                                                                                      
Use case: Multiple contracts indexing the same event type can direct their custom tables to a shared database, enabling streaming MVs to read from a single source table instead of UNION ALL across per-contract databases.

### DIFF
--- a/core/src/database/clickhouse/generate.rs
+++ b/core/src/database/clickhouse/generate.rs
@@ -49,7 +49,7 @@ pub fn generate_tables_for_indexer_clickhouse(
 
             // Generate custom tables if defined
             if let Some(tables) = &contract.tables {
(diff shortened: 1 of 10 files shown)


joshstevens19 (changes requested):
looks good but we need public docs and changelog please

joshstevens19 (approved):

PR #408, opened 2026-04-15
fix for no-code panic bug

masonhgn:
In no_code.rs line 361, the trace callback filters the incoming batch for TraceResult::NativeTransfer entries and calls .unwrap() on the result

native_transfer_block_consumer always fires trigger_event with a block only batch first before firing the NativeTransfer batch, so the filter produces an empty iterator and the unwrap causes a panic.

This seems to happen on the first block processed whenever native_transfers is enabled with no_code CSV output, making this portion unusable

I came across this bug when trying to verify my own version of a fix proposed in this PR #368 

update:

To add coverage for no_code.rs fix:

- Given that no_code.rs had no_code_callback taking NoCodeCallBackParams, it was difficult writing unit test coverage with this structure because NoCodeCallBackParams is a huge load of stuff that this unit test shouldn't be concerned with

- extracted the code this PR is concerned with into a function first_metadata() so it could be 
unit tested in isolation

- wrote 5 unit tests supporting each outcome:
1. first_metadata_event_returns_fields_from_first_entry (Event)
2. first_metadata_trace_native_transfer_returns_fields (NativeTransfer variant)
3. first_metadata_trace_block_returns_fields (Block variant) <----- this is where the bug was
4. first_metadata_empty_event_returns_none (empty Event input)
5. first_metadata_empty_trace_returns_none (empty Trace input)

### DIFF
--- a/core/src/event/callback_registry.rs
+++ b/core/src/event/callback_registry.rs
@@ -47,8 +47,7 @@ pub enum CallbackResult {
 }
 
 impl CallbackResult {
-    ///return the (from_block, to_block, network) entry regardless of variant
-
+    /// return the (from_block, to_block, network) entry regardless of variant
     pub fn first_metadata(&self) -> Option<(U64, U64, String)> {
         match self {
             Self::Event(events) => events.first().map(|e| {
--- a/core/src/event/callback_registry.rs
+++ b/core/src/event/callback_registry.rs
@@ -46,6 +46,33 @@ pub enum CallbackResult {
     Trace(Vec<TraceResult>),
 }
 
+impl CallbackResult {
+    ///return the (from_block, to_block, network) entry regardless of variant
+
+    pub fn first_metadata(&self) -> Option<(U64, U64, String)> {
+        match self {
+            Self::Event(events) => events.first().map(|e| {
+                (
+                    e.found_in_request.from_block,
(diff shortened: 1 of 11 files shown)


ExoMonk (commented):
Great one, could you just ensure this throuh adding E2E dedicated test ?

masonhgn:
> Great one, could you just ensure this throuh adding E2E dedicated test ?

sounds good thx, working on this

masonhgn:
@ExoMonk should I add separate unit tests to satisfy the codecov validation above or are the e2e tests sufficient? Totally can do both just wanted to avoid redundancy

ExoMonk:
> @ExoMonk should I add separate unit tests to satisfy the codecov validation above or are the e2e tests sufficient? Totally can do both just wanted to avoid redundancy

If you can add unit quickly for coverage it's best then we approve everything !

masonhgn:
added unit coverage + updated description @ExoMonk

ExoMonk (approved):
LGTM

masonhgn:
Saw clippy failure, seems to be just from a comment. pushing a fix in a bit

ExoMonk (approved):

joshstevens19 (approved):

PR #410, opened 2026-04-15
fix: ClickHouse String casting

ExoMonk:


JosepBove:
The escaping is incomplete. It handles ' but not \ . ClickHouse uses C-style backslash escaping, so an attacker-controlled string containing a trailing \ will escape the closing quote delimiter and merge with adjacent SQL. This is exploitable: e.g., col_a = \ + col_b = OR 1=1 -- turns a DELETE into ALTER TABLE t DELETE WHERE ... OR 1=1, wiping the entire table.

If you want to go with this approach you either escape backslashes first, then quotes or use parameterized queries

joshstevens19 (approved):

PR #420, opened 2026-04-21
Parallel historical backfill (rebase of #380 + fixes + tests)

zannis:
Rebases and finishes off @ExoMonk's [#380](https://github.com/joshstevens19/rindexer/pull/380) on top of the current reorg-handling master, fixes a handful of bugs the rebase surfaced, and adds unit + Docker-based e2e coverage for the parallel path.

Closes #380 (supersedes it).

## Summary

- Historical backfill can now split a block range into `fetch_concurrency` chunks fetched concurrently, with a reorder buffer guaranteeing strict block-order delivery to the consumer.
- Opt-in via a single top-level config field — `None` preserves current sequential behaviour for zero-config upgrades.
- Factory contracts and ranges < 1000 blocks still use the sequential path (not worth the overhead).

```yaml
config:
  fetch_concurrency: 4
```

## What's new on top of #380

**Rebased onto current master.** PR #380 branched before the reorg-handling refactor (#401) — `ChainProvider` trait, `FetchLogsResult.reorg`, `reorg_coordinator` threading, and `syncing_log()`/`completed_log()` helpers. The parallel path is wired through these new APIs; sequential behaviour is byte-identical to master.

**Bugs fixed during rebase:**
- **Worker panic leaked `active_workers` counter.** A panic between `fetch_add` (dispatcher) and the explicit `fetch_sub` (worker) left the counter stuck, eventually deadlocking the dispatcher once enough workers had panicked. `WorkerDropGuard` now single-sources cleanup — the Drop impl unconditionally releases the slot and notifies, so the path is panic-safe.
- **Off-by-one in `total_blocks`.** The old exclusive count treated a range `[a, a]` as 0 blocks; now inclusive via `saturating_add(1)`.
- **Theoretical u64 overflow** in the chunk-boundary arithmetic — guarded with `saturating_add`.
- **Silent drop on partial-batch send error.** The worker's partial-flush path `let _ = tx.send(...)` would keep burning RPC after the downstream reorder task had closed. Now breaks out of the loop on send error.
- **Coalescer stale block-range after reorg flush.** In live indexing, when a reorg batch arrived mid-coalesce, the code flushed the coalesced batch + the reorg batch, then `continue`d the inner loop with stale `final_from_block` / `final_to_block` variables. Subsequent batches' `min`/`max` arithmetic would poison the `from_block` passed to callbacks. Now uses a labelled `continue 'outer` so each coalescing cycle rebuilds its range from a fresh first batch.

**Simplifications:**
- `fetch_logs_once` is a pure function (no channel side-effects) extracted out of the sequential logic — this is what workers call.
- `ReorderBuffer` is a standalone unit-testable type with a 6-line async wrapper in the spawn site; dispatcher/reorder logic no longer inlined.
- `process.rs` consumer extracted a `dispatch_batch` helper, collapsing four near-identical spawn-or-push blocks.
- Hot-loop allocations (`info_log_name`, `block_clock`, `cached_provider`) hoisted out of the per-fetch loop in `parallel_worker`.

## How it works

1. **Dispatcher** slices `[from_block, snapshot_to_block]` into `chunk_size` pieces, each tagged with a monotonic `sequence_id`. Spawns workers up to `effective_concurrency`, gated by `AdaptiveConcurrency::current()` so errors can scale the pipeline down.
2. **Worker** calls `fetch_logs_once` in a loop over its sub-range, flushing partial batches at `MAX_WORKER_RESULTS=1000` to bound per-worker memory. Sends a final batch tagged `is_final=true` and — on any exit path, including panic — `WorkerDropGuard` guarantees the reorder buffer sees either the final batch or a synthetic error batch.
3. **Reorder buffer** holds incoming batches keyed by `sequence_id`. Emits results for the current `next_expected` immediately (partial or final), buffers the rest, and drains the buffered prefix once each final batch arrives.
4. **Consumer** in `process.rs` coalesces ready batches via `now_or_never()` on the mpsc stream for amortized callback overhead; reorg batches bypass coalescing because the coordinator's rollback requires standalone delivery.

## Usage

```yaml
config:
  fetch_concurrency: 8   # recommended 4-10 for dedicated RPC nodes; capped at 32
```

Chunking:
- `total_blocks < 1000`  → sequential path
- `factory_event`        → sequential path
- otherwise              → `effective_concurrency = min(concurrency, floor(total_blocks/1000))` workers of `chunk_size = max(1000, total_blocks/concurrency)` blocks each

## Tests

**Unit tests** — 21 new tests in `core/src/indexer/fetch_logs.rs::tests::parallel`:
- `plan_parallel_fetch` chunk/worker math (7 cases: small range, exact threshold, capped-at-32, zero concurrency, u64::MAX/2, ...)
- `ReorderBuffer` (6 cases: in-order, out-of-order, partial-then-final, reverse arrival, buffered-partial-then-final-flush-on-gap-close, empty-final-still-advances)
- `WorkerDropGuard` (4 cases: panic sends error batch, clean exit no-ops, closed-channel cancels pipeline, counter released even on channel-full path)
- `fetch_logs_once` (4 cases: empty range, past snapshot, logs advance past last, inverted range self-corrects)

**End-to-end tests** — `core/tests/e2e_parallel_fetch.rs` using testcontainers (Anvil + Postgres):
- `parallel_fetch_parity_across_worker_counts` — mines ~8550 blocks, emits 50 `Ping` events, runs rindexer's no-code pipeline three times at `fetch_concurrency = 1`, `4`, `8` (actually engaging 4 and 8 workers respectively), and asserts all three runs produce **byte-identical** rows across `(id, block_number, tx_hash, log_index, block_hash, sender)`.
- `parallel_fetch_large_range_eight_workers` — 5430-block range with 8 requested (engages 5), verifies block-order preservation and tx_hash integrity.

Run with:
```bash
cargo test -q -p rindexer --test e2e_parallel_fetch -- --ignored --test-threads=1
```

## Test plan

- [x] Existing unit suite still passes (595 tests, 4 ignored — same as master)
- [x] `cargo clippy -q --all-targets -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] New unit tests pass
- [x] Docker e2e tests pass locally (c=1 sequential + c=4 and c=8 parallel paths engaged, parity verified)
- [ ] Maintainer spot-check of the reorg-bypass semantics in `process.rs` (changed from `continue` → `continue 'outer` to rebuild block-range state cleanly)


ExoMonk (approved):
1 conflict but LGTM

PR #424, opened 2026-04-25
fix(block-timestamp): retry RPC fetch, refuse NULL writes, NOT NULL on schema

ExoMonk:
## Summary                                             

Fix against silent `block_timestamp = 0` writes when the upstream RPC fails to return a block header. 

## Root cause                                          

When `manifest.timestamps = false` (default), `eth_getLogs` responses without a `blockTimestamp` field flow through with `tx_metadata.block_timestamp = None`. The write path tried a per-block cache populated by `prefetch_block_timestamps`, which on RPC failure logged the error, returned `None`, and **never inserted the cache entry**. Downstream the column was silently skipped from the row, then the executor emitted `EthereumSqlTypeWrapper::Null`, which surfaced as `0` on the destination NUMERIC column.

### DIFF
--- a/core/src/database/clickhouse/generate.rs
+++ b/core/src/database/clickhouse/generate.rs
@@ -111,7 +111,7 @@ fn generate_event_table_clickhouse(abi_inputs: &[EventInfo], schema_name: &str) 
                     {}
                     tx_hash FixedString(66),
                     block_number UInt64,
-                    block_timestamp Nullable(DateTime('UTC')),
+                    block_timestamp DateTime('UTC'),
                     block_hash FixedString(66),
                     network String,
                     tx_index UInt64,
--- a/core/src/database/generate.rs
+++ b/core/src/database/generate.rs
@@ -39,7 +39,7 @@ fn generate_event_table_sql_with_comments(
                     {event_columns} \
                     tx_hash CHAR(66) NOT NULL, \
                     block_number NUMERIC NOT NULL, \
-                    block_timestamp TIMESTAMPTZ, \
+                    block_timestamp TIMESTAMPTZ NOT NULL, \
                     block_hash CHAR(66) NOT NULL, \
                     network VARCHAR(50) NOT NULL, \
                     tx_index NUMERIC NOT NULL, \
--- a/core/src/indexer/last_synced.rs
+++ b/core/src/indexer/last_synced.rs
@@ -249,6 +249,9 @@ async fn update_last_synced_block_number_for_file(
         if let Some(last_block_value) = last_block { to_block > last_block_value } else { true };
 
     if last_block.is_none() || to_block_higher_then_last_block {
+        if let Some(parent) = Path::new(&file_path).parent() {
+            fs::create_dir_all(parent).await?;
+        }
         let temp_file_path = format!("{file_path}.tmp");
 
         let mut file = File::create(&temp_file_path).await?;
--- a/core/src/indexer/tables.rs
+++ b/core/src/indexer/tables.rs
@@ -1430,6 +1430,33 @@ async fn get_cached_block_timestamp(network: &str, block_number: u64) -> Option<
     cache.get(&(network.to_string(), block_number)).copied()
 }
 
+/// Predicate: would the column-write path silently emit Null for a critical
+/// block_timestamp column? True when the row is missing the value AND no DDL
+/// default is configured. Narrow scope — only `block_timestamp` /
+/// `rindexer_block_timestamp`; other columns retain Null-fallback for legitimate
+/// optional-value cases (failed view calls, optional enrichment).
+fn block_ts_guard_trips(
+    column_name: &str,
+    column_default: Option<&String>,
+    row_has_value: bool,
+) -> bool {
+    if row_has_value || column_default.is_some() {
+        return false;
+    }
+    column_name == "block_timestamp" || column_name == "rindexer_block_timestamp"
+}
+
+/// Best-effort block-number hint for guard-trip error messages. Prefers the
+/// canonical `rindexer_block_number` injected key, falls back to user-aliased
+/// `block_number`, returns `<unknown>` when neither is present.
+fn extract_block_num_hint(columns: &HashMap<String, EthereumSqlTypeWrapper>) -> String {
+    columns
+        .get("rindexer_block_number")
+        .or_else(|| columns.get("block_number"))
+        .map(|v| format!("{:?}", v))
+        .unwrap_or_else(|| "<unknown>".to_string())
+}
+
 /// Transaction metadata available for table value references.
 #[derive(Clone, Debug)]
 pub struct TxMetadata {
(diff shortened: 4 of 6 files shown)


core/src/indexer/last_synced.rs
```diff
@@ -249,6 +249,9 @@ async fn update_last_synced_block_number_for_file(
         if let Some(last_block_value) = last_block { to_block > last_block_value } else { true };
 
     if last_block.is_none() || to_block_higher_then_last_block {
+        if let Some(parent) = Path::new(&file_path).parent() {
+            fs::create_dir_all(parent).await?;
+        }
```
zannis, line 254: is this intended?

core/src/database/clickhouse/generate.rs
```diff
@@ -111,7 +111,7 @@ fn generate_event_table_clickhouse(abi_inputs: &[EventInfo], schema_name: &str)
                     {}
                     tx_hash FixedString(66),
                     block_number UInt64,
-                    block_timestamp Nullable(DateTime('UTC')),
+                    block_timestamp DateTime('UTC'),
```
zannis, line 114: this will be a breaking change for existing indexers - just noting as we need to add to docs

core/src/indexer/tables.rs
```diff
@@ -1314,36 +1314,65 @@ async fn prefetch_block_timestamps(
...
+                const MAX_RETRIES: u32 = 3;
+                let backoffs = [
+                    std::time::Duration::from_millis(100),
+                    std::time::Duration::from_millis(500),
+                    std::time::Duration::from_secs(2),
+                ];
+                let mut attempt: u32 = 0;
+                loop {
+                    match provider_clone
+                        .get_block_by_number_batch_with_size(
```
zannis, line 1330: i like this but maybe worth moving this inside `get_block_by_number_batch_with_size`

zannis: also this adds 3 retries but if these fail for some reason then we are back to the original issue right?

core/src/indexer/tables.rs
```diff
@@ -1401,6 +1430,33 @@ async fn get_cached_block_timestamp(network: &str, block_number: u64) -> Option<
...
+
+/// Best-effort block-number hint for guard-trip error messages. Prefers the
+/// canonical `rindexer_block_number` injected key, falls back to user-aliased
+/// `block_number`, returns `<unknown>` when neither is present.
+fn extract_block_num_hint(columns: &HashMap<String, EthereumSqlTypeWrapper>) -> String {
+    columns
+        .get("rindexer_block_number")
+        .or_else(|| columns.get("block_number"))
+        .map(|v| format!("{:?}", v))
+        .unwrap_or_else(|| "<unknown>".to_string())
```
zannis: one of these 2 should always be there right? maybe avoid defaulting to "unknown" as it might break data

core/src/indexer/tables.rs
```diff
@@ -1314,36 +1314,65 @@ async fn prefetch_block_timestamps(
                 // Wait for backoff if rate limited (for free nodes)
                 ADAPTIVE_CONCURRENCY.wait_for_backoff().await;
```
zannis, line 1316: One concern about the scope of this layer: it's only consulted by `process_table_operations` (custom tables). The standard event-table writer in `no_code.rs:449/511` reads `result.tx_information.block_timestamp` directly and emits `EthereumSqlTypeWrapper::DateTimeNullable(...)` without ever touching `BLOCK_TIMESTAMP_CACHE` or `prefetch_block_timestamps`. So under the default config (`manifest.timestamps = false`) with an RPC that omits `blockTimestamp`, the retry never runs — the new NOT NULL schema turns the silent corruption into a hard insert error on fresh tables (good), but with no hydration on this path, `trigger_event` will retry the same broken batch indefinitely. Could the prefetch cache lookup be wired into the standard writer too, before `insert_bulk`?

core/src/database/generate.rs
```diff
@@ -39,7 +39,7 @@ fn generate_event_table_sql_with_comments(
                     {event_columns} \
                     tx_hash CHAR(66) NOT NULL, \
                     block_number NUMERIC NOT NULL, \
-                    block_timestamp TIMESTAMPTZ, \
+                    block_timestamp TIMESTAMPTZ NOT NULL, \
```
zannis, line 42: `CREATE TABLE IF NOT EXISTS` means this only applies to fresh tables. `migrations.rs` ends at `V1` which originally added the column **nullable**, so existing Postgres deployments keep the nullable column on upgrade and the silent-NULL bug remains unfixed for them — same shape as the ClickHouse breaking-change you flagged, but without even the loud error CH gives. Worth either a `Migration::V2` that backfills + `ALTER COLUMN ... SET NOT NULL`, or at minimum calling out the upgrade path in docs alongside the CH note.

core/src/indexer/tables.rs
ExoMonk: JsonRpcCachedProvider already wraps alloy::transports::layers::RetryBackoffLayer

The retry I added is on top of that, scoped to  this specific call site because the failure surfaced here in production despite the transport-level retry running.
 Pushing it into get_block_by_number_batch_with_size would amplify cost for callers that don't need it (e.g. attach_log_timestamps in blockclock/fetcher.rs:173-206 already has its own recursive retry). I'd prefer to keep it at the call site

On the failure mode: Layer 1 alone doesn't solve the bug;; Layers 2 + 3 are the load-bearing pieces. If all 3 attempts fail, execute_postgres_operation / execute_clickhouse_operation returns Err (Layer 2), and even if Layer 2 were bypassed, Postgres/ClickHouse rejects the NULL insert (Layer 3 NOT NULL). The outer callback_registry::trigger_event:530-575 retry loop with exp backoff (capped at 15s) then re-attempts the batch. Worst-case post-fix: indexer pauses + retries forever: o silent corruption

hane-polymarket:
https://github.com/joshstevens19/rindexer/pull/468 supersedes this pr.

Since the older PR was opened, Polymarket mitigated the production issue and repaired the affected history independently:

- https://github.com/Polymarket/polymarket-indexer/pull/90 enabled `timestamps: true`.
- The historical repair replaced the epoch timestamps, including 7,098,707 trade rows. The runbook still treats DENG-83 prevention as open and any new epoch-timestamp rows as a regression.
- https://github.com/Polymarket/polymarket-indexer/pull/489 now pins rindexer commit `e24b70c0`, which does not include the older PR; its staging soak found no missing timestamps.

Those steps fixed Polymarket’s deployed state, but they do not eliminate the upstream failure mode: if timestamp metadata is absent and cache/RPC resolution also fails, a custom-table operation can still write `NULL` or an epoch timestamp.

This PR addresses that remaining failure mode against current `master` with a narrower change. It detects timestamp dependency from the `$rindexer_block_timestamp` source reference, covering aliases and timestamp-dependent conditions; resolves and hydrates timestamps before any configured sink write; and, after the provider’s retry budget is exhausted, leaves progress unadvanced so the range can be retried safely without replaying committed sink writes or spinning forever in the callback retry loop. It avoids the older PR’s broader raw-schema, destination-name, and inner-retry changes.

PR #430, opened 2026-05-02
fix: add missing watch field to StartDetails in generated code and docs

mcamara:
## Summary
  - The `watch: bool` field was added to `StartDetails` (as part of hot-reload support) but the code template in `build.rs` that generates user Rust projects was not updated, any project scaffolded  after that change would fail to compile with a missing struct field error.
  - Fixed the two `StartDetails` examples in `indexers.mdx` that were also missing both watch and `cron_scheduler_handle`, these would have failed to compile if copy-pasted by users.
  - Hot-reload is documented as unsupported for Rust projects, so `watch: false` is the correct hardcoded value in both the generator template and the docs.

## Test Plan
  - Run `rindexer new rust` to scaffold a new Rust project and verify it compiles without errors
  - Confirm the generated `main.rs` contains `watch: false` in the `StartDetails` initializer

### DIFF
--- a/core/src/generator/build.rs
+++ b/core/src/generator/build.rs
@@ -579,6 +579,7 @@ serde = {{ version = "1.0", features = ["derive"] }}
                                 override_port: port,
                             },
                             cron_scheduler_handle: None,
+                            watch: false,
                         })
                         .await;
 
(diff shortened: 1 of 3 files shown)


joshstevens19 (approved):

mcamara:
Hello @joshstevens19, I've added some tests in `core/src/generator/build.rs` to cover the codecov warning.

joshstevens19 (approved):

Issue #348, opened 2025-11-19
[Feature] Add detection & warning when connected RPC is pruned or unable to serve historical logs


bh2smith:
## Summary
When users configure Rindexer with a `startBlock` that lies outside the history retained by their execution client, indexing may silently produce incomplete or incorrect results. Different clients behave differently when queried for pruned history — some return JSON-RPC errors (e.g., Erigon), while others return empty results (e.g., Reth), which makes it difficult for users to know that their node cannot serve the requested range.

Rindexer should detect this situation early and warn the user before indexing begins.

## Proposed Solution

On startup, before indexing begins, Rindexer should perform a history support probe against the configured RPC:
1. Identify the client (optional, for logging):
   - Call `web3_clientVersion`
2. Probe the configured `startBlock`:
  - Call `eth_getLogs` for a very small window: `fromBlock = startBlock, toBlock = startBlock + 10`
  - If this returns:
  - A RPC error (Erigon-style pruning errors) → treat as pruned
  - An empty array → ambiguous, continue to step 3
  - Normal logs → OK
3. If ambiguous, probe receipts:
  - Fetch the block via `eth_getBlockByNumber`
  - If it contains transactions, call `eth_getTransactionReceipt` on one of them
  - If receipts return error/null/inconsistent data → treat as pruned
4. If the node appears pruned, Rindexer should:
  - Print a clear warning and/or refuse to continue, and/or
  - Allow an override flag like `--ignore-prune-warnings`

This approach works consistently across Erigon, Reth, Geth, and others, and avoids relying on non-standard client-specific introspection.

PR #431, opened 2026-05-07
feat: add pruned RPC detection on startup (#348)

0xvectr:
Detect when RPC endpoints are pruned and cannot serve historical data from the configured start_block. On startup, each network's RPC is probed by fetching the start block and verifying transaction receipts.

- Add --ignore-prune-warnings CLI flag to bypass check
- Support RINDEXER_IGNORE_PRUNE_WARNINGS env var
- Fail with clear error message suggesting archive node or start_block adjustment
- Add prune_detection module with unit tests
- Update changelog and CLI documentation


0xvectr:
@bh2smith , @joshstevens19, pr to fix #348 
Could you review?

joshstevens19 (changes requested):
im not fully sure we should do this because it slows the startup down a lot...

PR #432, opened 2026-05-16
fix: task_tracker underflow

BitWonka:
gm 

When shutting down gracefully, it goes on infinite loop due to bloom skipped live blocks decrementing the counter without starting a task.

```
12:39:37  INFO 18446744073709551609 active indexing tasks pending.. shutting them down gracefully
12:39:37  INFO 18446744073709551609 active indexing tasks pending.. shutting them down gracefully
```

### DIFF
--- a/core/src/indexer/process.rs
+++ b/core/src/indexer/process.rs
@@ -708,12 +708,7 @@ async fn live_indexing_for_contract_event_dependencies(
                     .lock()
                     .await = ordering_live_indexing_details;
 
-                update_progress_and_last_synced_task(
-                    Arc::clone(config),
-                    to_block,
-                    indexing_event_processed,
-                )
-                .await;
+                update_progress_and_last_synced_task(Arc::clone(config), to_block, || {}).await;
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


core/src/indexer/process.rs
```diff
@@ -708,12 +708,7 @@ async fn live_indexing_for_contract_event_dependencies(
                     .lock()
                     .await = ordering_live_indexing_details;
 
-                update_progress_and_last_synced_task(
-                    Arc::clone(config),
-                    to_block,
-                    indexing_event_processed,
-                )
-                .await;
+                update_progress_and_last_synced_task(Arc::clone(config), to_block, || {}).await;
```
joshstevens19: why change this line?

joshstevens19 (changes requested):

core/src/indexer/process.rs
BitWonka: this line is the actual root cause of the underflow, the other file change is a defense to add a warning if something like this was triggered. 

1. `update_progress_and_last_synced_task` always calls `on_complete` once. 
2. This branch passed `indexing_event_processed` (a decrement), but continues before `trigger_event`, so the matching `indexing_event_processing()` increment never ran. 
3. skipped bloom blocks decremented with no increment => the counter underflows. no task starts here, so it shouldn't touch the counter, hence the no-op || {}.

joshstevens19 (approved):

PR #440, opened 2026-06-22
Stabilize Kafka reorg stream test

SkandaBhat:
## Summary
- Explicitly create the Kafka topic in the Kafka reorg stream integration test before subscribing.
- Keep the change scoped to the test harness; production stream behavior remains unchanged.

## Why
The test was relying on Kafka auto-topic creation. In ephemeral Kafka containers that is racy: the consumer can subscribe or poll before the broker has materialized the topic, which intermittently returns `UnknownTopicOrPartition`. Creating the topic with the Kafka admin client makes the fixture deterministic and preserves the existing test intent.

## Validation
- `cargo fmt --check`
- 10x: `cargo nextest run -p rindexer --all-features --test e2e_reorg_streams stream_reorg_reaches_kafka`
- 5x: `cargo test -p rindexer --features kafka --test e2e_reorg_streams stream_reorg_reaches_kafka`
- Fresh branch sanity: `cargo nextest run -p rindexer --all-features --test e2e_reorg_streams stream_reorg_reaches_kafka`

### DIFF
--- a/core/tests/e2e_reorg_streams.rs
+++ b/core/tests/e2e_reorg_streams.rs
@@ -283,6 +283,32 @@ async fn stream_reorg_reaches_rabbitmq() {
 // ---------------------------------------------------------------------------
 
 #[cfg(feature = "kafka")]
+async fn create_kafka_topic(bootstrap: &str, topic: &str) {
+    use rdkafka::{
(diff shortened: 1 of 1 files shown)


SkandaBhat:
Superseded by #439, which now includes this Kafka test stabilization as a separate commit.

PR #441, opened 2026-06-22
fix(reorg): quote reserved-word identifiers in reversal SQL; stabilize Kafka reorg stream test

zannis-poly:
## Summary

Two independent fixes surfaced while investigating reorg-test CI failures:

### 1. `fix(reorg)`: quote identifiers in reversal SQL (real correctness bug)
The reorg derived-table reversal interpolated user-defined column names **unquoted** into the generated SQL. A derived table keyed on a reserved word — e.g. an ERC20 aggregation with `where: { holder: "$to" }` — produced `SELECT to … GROUP BY to`, which Postgres rejects with `syntax error at or near "to"`. The rollback then failed every retry and the live-reorg recovery test timed out.

- Identifiers are now quoted per backend (Postgres double-quotes, ClickHouse backticks) across the snapshot `SELECT`/`GROUP BY`, the PG apply-phase join, and the CH `Join` keys / `joinGet` / exists-filter.
- The pre-existing `quote_identifier` helper relies on a 7-word reserved list that omits `to`/`from`, so it would not have caught this — hence always-quoting here.
- `set_clauses` are intentionally left unquoted: the ClickHouse set-clause rewriter parses bare `dt.`/`snap.` tokens.
- New regression test `test_reorg_reversal_reserved_word_column` (real Postgres via testcontainers) reproduces the exact error and passes after the fix. Full `e2e_reorg` suite is green, including the ClickHouse reversal path.

### 2. `test(streams)`: stabilize the flaky Kafka reorg-stream test
`stream_reorg_reaches_kafka` intermittently failed with `UnknownTopicOrPartition`. The test relied on lazy topic auto-creation and a consumer that treated the first stream item as fatal.

- Pre-create the topic via `AdminClient` before subscribing/producing (mirrors the RabbitMQ test declaring its topology up front).
- A freshly-created topic can still briefly surface transient metadata errors to the consumer, so the consume loop now tolerates them and keeps polling within the timeout instead of panicking on the first error event.

## Testing
- `cargo nextest run -p rindexer --test e2e_reorg` — 30/30 pass (incl. `test_reorg_clickhouse_add_reversal`)
- `cargo nextest run -p rindexer --features kafka --test e2e_reorg_streams stream_reorg_reaches_kafka` — pass
- fmt + clippy clean

### DIFF
--- a/core/src/indexer/reorg/task.rs
+++ b/core/src/indexer/reorg/task.rs
@@ -226,6 +226,19 @@ struct ReversalSnapshot {
     set_clauses: Vec<String>,
 }
 
+/// Quote a SQL identifier for Postgres (double quotes), escaping any embedded
+/// quotes. Required because user-defined event/derived columns can collide with
+/// SQL reserved words (e.g. `to`, `from`).
+fn quote_pg_ident(name: &str) -> String {
+    format!("\"{}\"", name.replace('"', "\"\""))
+}
+
+/// Quote a SQL identifier for ClickHouse (backticks), escaping any embedded
+/// backticks. Mirrors `quote_pg_ident` for the ClickHouse backend.
+fn quote_ch_ident(name: &str) -> String {
+    format!("`{}`", name.replace('`', "\\`"))
+}
+
 impl ReorgTask {
     /// Returns ` AND network = '<network>'` when not cross-chain, empty string otherwise.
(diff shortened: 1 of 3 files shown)


zannis:
Superseded by #442 (reopened from the correct account, plus codex-review hardening).

PR #451, opened 2026-07-19
fix: atomic insert

ExoMonk:
## fix: atomic [batch + cursor] commit for the no-code Postgres path

Closes the double-index race documented in `trigger_event`'s TODO: event batches
(COPY/multi-row INSERT) and the `rindexer_internal.last_synced_block` cursor were
committed separately (cursor async on another connection, errors swallowed), so any
crash/restart between the two re-fetched and re-inserted committed ranges.
Reproduced deterministically: cursor rollback + restart on a 300-block mainnet
window → 17,945 + 19,327 byte-identical duplicate `(tx_hash, log_index)` rows.

### Changes
- `client.rs`: `insert_bulk_with_cursor()` — one transaction for [COPY-or-INSERT
  batch + monotonic cursor UPDATE]; `warn!` when the cursor row matches 0 rows.
- `no_code.rs`: PG branch uses it (cursor = fetched `to_block`), gated by
  `atomic_pg_cursor` to Postgres-as-sole-raw-sink; block ordered after table
  operations (cursor must be the last retry-on-error write); table-ops shutdown
  checkpoint suppressed under the atomic path (it advances the same cursor row
  and would skip un-inserted raw rows on SIGTERM).
- `process.rs`: TODO updated with the remaining non-atomic paths.

### Semantics
Exactly-once for PG raw events under a single writer with callback concurrency 1
(the default). Dual-sink (PG+CH/CSV), factory streams, traces and streams keep
legacy at-least-once semantics — preconditions documented in code

### Validation
Hermetic E2E on real mainnet blocks: 5× `kill -9` during indexing → **0 duplicates,
0 loss** (distinct keys equal to un-killed control: 50,182 / 37,5
under the same kill-loop reproduces duplicates. Two independent adversarial reviews;
all blocking findings addressed in-diff.


core/src/indexer/no_code.rs
```diff
@@ -621,11 +625,113 @@ fn no_code_callback(params: Arc<NoCodeCallbackParams>) -> EventCallbacks {
...
+            // Atomic-cursor eligibility: ONLY when Postgres is the SOLE raw-event
+            // sink. With ClickHouse/CSV alongside, committing the cursor at the PG
+            // write would turn their crash-recovery from at-least-once into silent
+            // at-most-once (a committed cursor skips the range on restart before
+            // the other sinks ever saw it) — those configs keep the legacy path
+            // (plain insert_bulk; the async task advances the cursor only after
+            // the whole callback succeeds). The exactly-once contract additionally
+            // requires a SINGLE writer process and effective callback concurrency 1
+            // (the default; see trigger_event in process.rs).
+            let atomic_pg_cursor = params.store_raw_events
```
joshstevens19, line 653: **Blocking — factory events must be excluded from the atomic path.**

This gate has no factory exclusion, but factory child-address bookkeeping runs only *after* the callback returns (`core/src/event/config.rs:196-199`: `trigger_event(...)` then `update_known_factory_deployed_addresses(...)`).

Sequence with a PG-only factory config (e.g. `PairCreated` in `include_events`):
1. Batch with discovery logs hits this atomic path → `insert_bulk_with_cursor` commits rows **and** cursor.
2. Process crashes (kill -9 / OOM) before `update_known_factory_deployed_addresses` persists the child addresses.
3. On restart, resume reads the committed cursor (`start.rs:144-152`) → the discovery range is **never refetched**, the child addresses are never registered (they're only recovered from the persisted store, never re-derived from raw rows), and every event those child contracts ever emit is silently missed forever.

Pre-PR the cursor advanced only after the bookkeeping, so a crash there merely caused a refetch + duplicates. This turns that recoverable window into permanent loss. The `process.rs` comment acknowledges factory streams as a residual race, but there's no code guard here — this gate needs `&& !<is factory event>`, keeping factory events on the legacy path.

core/src/generator/events_bindings.rs
```diff
@@ -899,24 +901,74 @@ use super::super::super::typings::{indexer_name_formatted}::events::{handler_reg
...
-                        rindexer_error!("{event_type_name}::{handler_name} inserting bulk data: {{:?}}", e);
-                        return Err(e.to_string());
-                    }}
+                    {insert_call}
                 "#,
-                table_name =
-                    generate_event_table_full_name(indexer_name, &contract.name, &event.name),
-                handler_name = event.name,
-                event_type_name = event_type_name,
+                insert_call = if storage.postgres_enabled() && !storage.csv_enabled() {
```
joshstevens19: **Blocking — same factory crash-recovery race as the no-code path, in generated Rust handlers.**

Factory contracts are synthesized as real contracts with `include_events = [factory event]` (`manifest/core.rs:151-247`), so codegen emits the atomic `insert_bulk_with_cursor` handler for the factory's discovery event too — and this gate (`postgres_enabled() && !csv_enabled()`) has no factory exclusion. `FactoryEventProcessingConfig` persists child addresses only after the registry callback (`event/config.rs:196-199`), and restart resumes from the same `rindexer_internal` cursor the generated handler advances.

A crash between the generated handler's atomic commit and the address persist permanently skips the discovery range on restart → child addresses lost forever (see the companion comment on the `no_code.rs` gate for the full trace). Factory events need to be excluded here as well — e.g. keep the legacy `insert_bulk` template when the contract is a synthesized factory.

joshstevens19 (commented):
Reviewed this in depth (multi-angle review, findings cross-checked with a second independent model — both agreed on all of these). The atomic [batch + cursor] commit itself is sound and fixes a real bug, but there are **3 blocking issues** where moving the point-of-no-return earlier converts recoverable duplicate-work windows into permanent silent loss. Inline comments below with details and suggested fixes:

1. **Factory events must be excluded from the atomic path** (both the no-code gate and the codegen gate) — a crash after the atomic commit but before `update_known_factory_deployed_addresses` permanently loses child contracts.
2. **The table-ops reorder must be gated to `atomic_pg_cursor`** — under the infinite callback retry, a transient raw-insert failure now duplicates Insert-op rows and double-applies Add/Subtract arithmetic upserts, including in ClickHouse/CSV configs that gain nothing from the reorder.
3. **`callback_concurrency` must be part of the eligibility check** — with concurrency > 1 the monotonic cursor can commit past an uncommitted earlier batch, turning a crash into a permanent gap.

Non-blocking but worth doing while in here: `insert_bulk_with_cursor` duplicates ~60 lines of `insert_bulk`'s COPY/INSERT machinery and the copies already diverge in this PR (the new COPY error path swallows the `finish()` error via `let _ = writer.finish().await;` while this same PR fixes the original to propagate it) — making `bulk_insert_via_copy`/`bulk_insert_via_query` generic over `&impl tokio_postgres::GenericClient` would remove both the duplication and the divergence.

core/src/indexer/no_code.rs
```diff
@@ -621,11 +625,113 @@ fn no_code_callback(params: Arc<NoCodeCallbackParams>) -> EventCallbacks {
...
+                && params.postgres.is_some()
+                && params.clickhouse.is_none()
+                && params.csv.is_none();
+
+            // Process table operations
+            if !params.tables.is_empty() && !table_events_data.is_empty() {
```
joshstevens19: **Blocking — this reorder should be gated to `atomic_pg_cursor`, not applied to all sink configs.**

Table ops now run *before* the raw insert unconditionally, and `callback_registry.rs:539-566` retries the whole callback forever on `Err`. So any transient raw-insert failure (PG connection blip during the COPY, ClickHouse/CSV error in legacy configs) now re-runs `process_table_operations` on every retry:

- **Insert-type ops** append duplicate rows — no dedup by design (`tables.rs:3915-3927`), plain binary COPY with no ON CONFLICT (`batch_operations/dynamic.rs:240-268`).
- **Add/Subtract arithmetic upserts** double-apply their deltas — `col = COALESCE(col,0) + EXCLUDED.col` (`query_builder.rs:305-315`), and the sequence guard is deliberately dropped when arithmetic columns exist (`dynamic.rs:278,351`), so replays are not idempotent.

Pre-PR, a transient raw-insert failure retried cleanly (table ops hadn't run yet in that attempt). This replay was already possible via crash-restart, but the reorder widens it to *every* transient raw-insert error — including ClickHouse/CSV configs that gain nothing from the new ordering. The comment's "downstream routing PKs dedup" claim doesn't hold for Insert ops or arithmetic columns.

Suggested fix: keep the old ordering (raw insert first) for the legacy path and only run table-ops-first when `atomic_pg_cursor` is true.

core/src/indexer/process.rs
```diff
@@ -870,8 +870,15 @@ async fn trigger_event(
...
+        // Double-index race NOTE: for the no-code POSTGRES-sole-sink path this is
+        // closed — no_code_callback commits [batch + rindexer_internal cursor] in
+        // ONE transaction (insert_bulk_with_cursor), so this async advance is only
+        // the empty-range / tail bump (monotonic guard makes it a no-op otherwise).
+        // The race window still exists for: ClickHouse/CSV(-alongside-PG) configs
+        // (deliberately kept legacy — see atomic_pg_cursor in no_code.rs), custom
+        // Rust handlers, the trace path (its outer advance ignores the callback
+        // result), and factory streams (child-address bookkeeping commits after
+        // the callback — see DENG-414 review F2).
         update_progress_and_last_synced_task(config, to_block, indexing_event_processed).await;
```
joshstevens19, line 885: **Blocking (one-line fix) — the concurrency-1 precondition is documented but never enforced.**

The exactly-once contract assumes effective callback concurrency 1, but nothing checks it: `callback_concurrency` is user-configurable (`manifest/config.rs:11-16`) and applied at `process.rs:96-100` unless `index_event_in_order` is set. With `callback_concurrency: 2`, batch B2 (blocks 200-300) can atomically commit cursor=300 while B1 (blocks 100-200) is still uncommitted — the cursor UPDATE is only monotonic (`$1 > last_synced_block`), not ordering-aware. A crash before B1 commits means restart resumes from 300 and **blocks 100-200 are permanently skipped** — silent at-most-once, strictly worse than the duplicates this PR fixes.

Suggested fix: include the effective concurrency in the `atomic_pg_cursor` eligibility check in `no_code.rs` (and mirror it in the codegen gate), so the unsafe combination falls back to the legacy at-least-once path.

ExoMonk: - Concurrency: atomic path now requires index_event_in_order for the event OR
  callback_concurrency <= 1, in both no_code and the generated handlers.

core/src/indexer/no_code.rs
ExoMonk: Factory events: excluded from the atomic path in both no-code and codegen. The gate
mirrors is_factory_filter_event (all details are factory-discovery for this event).
Verified on the factory example: the PoolCreated discovery event emits the legacy
path, plain factory-address contracts and child contracts emit the atomic path.

core/src/generator/events_bindings.rs
ExoMonk: Factory events: excluded from the atomic path in both no-code and codegen. The gate
mirrors is_factory_filter_event (all details are factory-discovery for this event).
Verified on the factory example: the PoolCreated discovery event emits the legacy
path, plain factory-address contracts and child contracts emit the atomic path.

core/src/indexer/no_code.rs
ExoMonk: Table-ops ordering: restored to two arms. The legacy arm (ClickHouse/CSV/factory/
concurrency>1) keeps the stock order — raw sinks first, then table ops with the
shutdown checkpoint. Only the atomic arm runs table ops before the raw insert, with
the checkpoint suppressed since the atomic cursor supersedes it — a transient raw
failure there retries against rolled-back state, so Insert ops and arithmetic
upserts can't double-apply. Covered by the arm-matrix codegen tests in this PR.

core/src/indexer/no_code.rs
```diff
@@ -625,93 +641,201 @@ fn no_code_callback(params: Arc<NoCodeCallbackParams>) -> EventCallbacks {
...
+            //   raw rows.
+            //   LEGACY arm (CH/CSV alongside PG, factory discovery events,
+            //   concurrency > 1, or no PG): the ORIGINAL ordering — raw sinks
+            //   first, table ops after — so a transient raw-sink failure retries
+            //   BEFORE table ops ran. Insert-type table ops are append-only (no
+            //   dedup) and arithmetic upserts are not idempotent, so re-running
+            //   them on every callback retry would duplicate/double-apply.
+            //   Cursor advance stays with the async task after the callback
+            //   succeeds (at-least-once, pre-existing semantics).
+            // Streams/chat run after either arm and deliberately never propagate
```
skanda-polymarket, line 696: should streams/chat also keep the legacy path? the cursor commits before either runs below, so a crash in that window skips those messages on restart. i think this needs an outbox before it can be atomic.

core/src/generator/events_bindings.rs
```diff
@@ -899,24 +902,91 @@ use super::super::super::typings::{indexer_name_formatted}::events::{handler_reg
...
+                    // the no-code path (insert_bulk_with_cursor): closes the
+                    // double-index race for Rust-mode generated storage handlers.
+                    // Postgres is the sole raw sink here by construction (the
+                    // generated context selects postgres OR clickhouse, never both);
+                    // csv-enabled projects keep the legacy path (cursor must not
+                    // commit before the csv append). Cursor advances to the batch's
+                    // max log block — the log-free tail refetches empty on restart
+                    // and the async task still bumps to the fetched to_block.
+                    format!(
+                        r#"let cursor = rindexer::BulkCursorUpdate {{
```
skanda-polymarket: i dont think generated rust handlers are safe to opt into this automatically. any custom logic after this commit is outside the transaction, and a crash before the handler returns makes restart skip it. probably keep rust handlers legacy unless the user explicitly opts into raw-storage-only behavior.

core/src/database/postgres/client.rs
```diff
@@ -456,6 +464,133 @@ impl PostgresClient {
...
+        );
+        let cursor_rows = transaction
+            .execute(
+                &cursor_query,
+                &[&EthereumSqlTypeWrapper::U64(cursor.to_block), &cursor.network],
+            )
+            .await
+            .map_err(|e| e.to_string())?;
+
+        transaction.commit().await.map_err(|e| e.to_string())?;
```
skanda-polymarket: `cursor_rows == 0` should probably error before commit. right now we commit the event rows and return `Ok`, even though the cursor did not move. that breaks the batch + cursor contract and can duplicate on restart.

core/src/indexer/no_code.rs
skanda-polymarket: i think this still has the same retry issue inside the atomic arm. table ops commit separately, so if the raw insert fails after this the table ops run again on retry. this isnt rolled back by `insert_bulk_with_cursor`; worth keeping table events on the legacy path until both share a transaction.

core/src/indexer/process.rs
skanda-polymarket: this is checked at codegen time, but callback concurrency comes from the yaml at runtime. changing it to >1 without regenerating leaves the handler atomic and reopens the out-of-order cursor gap. can we validate this at runtime?

ExoMonk: Resolved by the rust-handler change above — there's no generated atomic path left to
drift when the yaml changes. The no-code gate computes concurrency from the manifest
the indexer is actually running with, so the check can't go stale there.

ExoMonk: cf
no_code.rs:643 (table ops):
True, and it's a deliberate tradeoff documented in the ordering contract now: table
ops keep their own transactions, so a failure between them and the atomic commit
re-runs them on retry — the same at-least-once exposure they have on the legacy path
today (callback retry re-runs them in stock rindexer too). Running them after the
atomic commit instead would turn a crash in between into a permanent skip, which is
strictly worse. Folding them into the same transaction is the real fix and is
exactly the direction of the sync_together block sink in #453.

core/src/generator/events_bindings.rs
ExoMonk: Agreed — generated handlers now always emit the legacy insert. Users own these files
and add logic inside them, so exactly-once storage is a per-handler opt-in via the
public insert_bulk_with_cursor API; qualifying scaffolds (Postgres sole sink,
non-factory) get a hint comment pointing at it, with the callback_concurrency <= 1
and no-side-effects-after-the-call requirements spelled out.

core/src/indexer/no_code.rs
ExoMonk: Agreed — done. The atomic gate now also requires no streams and no chat clients, so
those configs keep the legacy at-least-once path (crash → refetch → re-dispatch).
An outbox would let them go exactly-once too, but that's the sync_together/block-sink
scale of machinery — out of scope here.

core/src/indexer/no_code.rs
skanda-polymarket: can we keep table events on the legacy path for now and track the shared transaction separately?

skanda-polymarket (approved):

joshstevens19 (approved):

joshstevens19 (approved):

PR #453, opened 2026-07-20
feat(sync_together): per-block atomic lockstep live indexing for grouped events

joshstevens19:
## What

Adds a `sync_together` manifest key for no-code + Postgres projects: named groups of events that, during **live indexing only**, advance block-by-block in lockstep — for each block, every grouped event's raw event rows, custom-table operations, reorg journal, and `last_synced_block` checkpoints commit in **one Postgres transaction per block, per network**. Readers never observe a block partially applied across the group, and checkpoints can never drift from data. Historical backfill is unchanged.

Two YAML entry points:

```yaml
# cross-contract groups
sync_together:
  - group: market-sync
    contracts:
      - name: CTFExchange
        events: [OrderFilled]
      - name: ConditionalTokens
        events: [PositionSplit]
```

```yaml
# or per-table opt-in (sugar for grouping that table's events)
tables:
  - name: total_balances
    cross_chain: true
    sync_together: true
```

Custom tables are included automatically when an event is grouped (the whole callback buffers). Transactions are strictly per network — never cross-chain, even for `cross_chain: true` tables. `start_block` is optional for grouped members: without one they start at the tip on first boot and, unlike ungrouped events, resume from their checkpoint on restarts.

## How

- **Buffer-and-flush sink** (`database/postgres/block_sink.rs`): grouped callbacks buffer `(sql, params)` / COPY / batch ops into a per-group sink (installed via a tokio task-local around `trigger_event_once`, so backfill stays on the eager path untouched); `PostgresClient::flush_sync_together_block` replays them in one tx with in-transaction checkpoint checks that (a) detect reorg rewinds performed by other loops and (b) skip members already committed past a block — closing the known hist→live double-index race for grouped events.
- **Lockstep loop** (`indexer/sync_together.rs`): one loop per group × network, modeled on the dependency-events live loop; fast-forwards stragglers to a common block, then commits block-by-block; shares the network's `ReorgCoordinator` and aborts in-flight windows via a new rollback epoch when a rollback interleaves.
- **Side effects** (streams/chat/CSV) defer until after the block commit: never published for uncommitted blocks, retries can't double-publish; at-most-once if the process dies between commit and dispatch (documented).
- **Validation**: 12 startup rules (no-code+Postgres only, ClickHouse rejected, factories rejected both directions, cron tables rejected, network/reorg congruence, dependency_events mutual exclusion incl. relationship-derived trees, reserved group-name prefix, native-transfers table flag rejected, ...).

## Also ships two general fixes

- **Reorg checkpoint rewinds use `LEAST()`** — a rollback can never move a still-backfilling event's checkpoint forward (affects all projects).
- **Upsert `subtract` creates rows at `-value`** — previously a row created by a subtract op (e.g. a holder first seen as a sender) got `+value`, corrupting subtract-based aggregates; `if:` conditions still see the positive event value; applies to no-code tables and typed-project batch ops. Tables built with subtract upserts may need a re-sync (changelogged).

## Docs / example

- New `yaml-config/sync-together` docs page + updates across 9 existing pages + changelog entries.
- Runnable example `examples/tables_sync_together` (cross-chain USDC balances on ethereum/arbitrum/optimism) — verified live: 3k+ real transfers indexed in lockstep with `SUM(balances) == credits - debits` holding exactly.

## Testing

- Unit: manifest serde/desugar, `plan_block_buckets`, upsert SQL shape (incl. subtract negation + deterministic conflict-key `ORDER BY` for cross-network deadlock avoidance).
- `core/tests/yaml_sync_together.rs` — 23 validation-rule tests.
- `core/tests/sync_together_flush.rs` — 8 flush-transaction tests on real Postgres (atomic commit/rollback, rewind detection, seam idempotency, COPY-in-tx, LEAST guard).
- `core/tests/tables_upsert_arithmetic.rs` — 4 subtract-semantics tests on real Postgres.
- `core/tests/e2e_sync_together.rs` — anvil e2e: paired same-block events across two contracts with a concurrent poller asserting no torn blocks; checkpoint equality; no duplicates.
- `core/tests/example_tables_sync_together.rs` — example manifest validation guard + `#[ignore]`d live smoke test.
- Regression: full lib suite, all 31 `e2e_reorg` tests, `e2e_hist_to_live_dup`, typed example projects compile, typings regeneration round-trip verified, `cargo fmt`/`clippy -D warnings` clean, docs build clean.


hane-polymarket (commented):
Found **two new P2 (High) issues** in head `d1965d7c`:

1. **Reorg recovery can permanently skip deleted blocks.** [start.rs:1309](https://github.com/joshstevens19/rindexer/pull/453/files) uses the restart resume point as the rewind floor when `start_block` is omitted. If the group resumes at 101 and another loop rolls checkpoints back to 98, the flush clamps that checkpoint to 100. Replay starts at 101, permanently omitting blocks 99–100. Distinguish never-indexed checkpoints from real rewinds instead of flooring both.
2. **Factory discovery misses children created during downtime.** [start.rs:632–634](https://github.com/joshstevens19/rindexer/pull/453/files) enables checkpoint-based resumption only for grouped events. Without `start_block`, grouped children resume their checkpoints, while synthesized factory discovery jumps to the current tip. Children deployed during downtime never enter the address filter, so their events are permanently missed. Discovery dependencies also need checkpoint-based resumption.

PR #455, opened 2026-07-21
feat: embedded-mode support and multi-url rpc failover with health-gated endpoints

joshstevens19:
## What

Additive support for embedding rindexer as a library inside a host process (built for [rflow](https://github.com/joshstevens19/rflow), useful to any embedder), plus multi-url RPC failover for indexing. Each section is independently useful.

### 1. Embedded-mode support
- `StartDetails.manage_process_signals: bool` — when `false`, rindexer skips its own SIGINT/SIGTERM handlers (which call `std::process::exit(0)`), letting the host own process lifecycle. `true` preserves today's behavior byte-for-byte.
- `StartDetails.disable_health_server: bool` — skip binding the health server when the host serves its own.
- Re-export the generic log-decoding machinery (`parse_log`, `ParsedLog`, `LogParam`, wrapper→JSON mappers) so embedders decode events in custom callbacks with exact stream-payload parity.
- Compile the provider mock unconditionally so embedders can test registry construction chain-free.

### 2. Multi-url RPC manifest
`Network.rpc` accepts `String | Vec<String>` via `RpcUrls` (single url still serializes to the plain string — full back-compat, round-trip tested). `create_client` takes the url slice; codegen templates and in-repo generated examples updated.

### 3. Health-gated endpoint failover for indexing
HTTP providers route through a sticky sequential failover transport with per-endpoint health gates: consecutive-error cooldowns, a ~10s tip prober marking endpoints degraded when they lag the best tip past the chain's reorg-safe distance, lazy per-endpoint chain-id verification (boot survives a dead secondary; wrong-chain endpoints are excluded), and transport retries capped to single digits so failures actually surface to failover. **Deliberate deviation from alloy's `FallbackLayer`**: its hedged parallel dispatch lets a lagging endpoint *win* tip reads — exactly the spurious-reorg hazard below — and multiplies CU spend on getLogs volume; the sticky sequential design is documented in the module header.

### 4. Endpoint switches near the tip (the hard part)
Failing over to a laggier endpoint must not feed the reorg coordinator different-but-canonical hashes and trigger a spurious rollback. On a switch (checked before **and immediately after** `get_latest_block`, so a failover triggered by that very call is caught) the live loop suspends parent-hash validation at a watermark: tips at or below it are ignored, the first tip past it refetches the window from the new endpoint — all-match resumes, divergence below the watermark rolls back as a **real** reorg, refetch-impossible rebuilds the window without rolling back. Block-range shrink state resets per switch. Tested at coordinator level and through the real live-indexing stream with scripted providers (switch-to-lagging fires no rollback; a genuinely orphaned window still rolls back from the divergence point).

### 5. Host surfacing
`ChainProvider::rpc_health() -> Vec<EndpointHealthSnapshot>`, `endpoint_switches()`, and a `StartDetails.rpc_endpoint_events` broadcast sink (`Switched`/`Degraded`/`Recovered`) so hosts can alert on failover. All urls and error strings are credential-redacted upstream before they reach snapshots, events, or logs.

No behavior changes for existing users; single-url configs behave exactly as before and all new fields default to current behavior.

## Verification (CI parity, at the branch tip)

- `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean
- `cargo nextest run --exclude rindexer_rust_playground --workspace` green (default and `--all-features`), including the docker-backed reorg/parallel-fetch integration tests and the new failover/reorg-suspension suites
- Consumed downstream by rflow's 34-scenario anvil e2e suite via the pinned rev
- Cross-reviewed before landing (independent review passes caught and fixed: a mid-iteration switch race in the reorg suspension, and credential leakage through raw transport error strings)


core/src/endpoint_failover.rs
```diff
@@ -0,0 +1,1495 @@
...
+        let mut last_error: Option<TransportError> = None;
+        let mut switch_reason: Option<String> = None;
+
+        for index in order {
+            let health = &state.endpoints[index];
+
+            // Lazily verify the endpoint's chain id before its first use so a
+            // wrong-chain endpoint never serves data and a down fallback
+            // never blocks boot.
+            if health.chain_status() == CHAIN_UNVERIFIED {
```
hane-polymarket, line 664: **[P1] Re-check chain mismatch before dispatch**

`candidate_order()` may include an unverified fallback, then a concurrent request or background probe can mark it `CHAIN_MISMATCH` before this check executes. Because the condition is then false, execution falls through to `oneshot` and may return data from the wrong chain. Re-check for `CHAIN_MISMATCH` immediately before dispatch, or serialize verification so only `CHAIN_OK` endpoints can serve requests.

core/src/indexer/fetch_logs.rs
```diff
@@ -1201,8 +1272,26 @@ async fn live_indexing_stream(
...
         let latest_block = cached_provider.get_latest_block().await;
+
+        // The call above can itself trigger the failover (the request died on
+        // the old endpoint and was answered by a replacement), in which case
+        // the block below already comes from the NEW endpoint. Re-check the
+        // switch counter BEFORE any reorg validation - otherwise a lagging
+        // replacement's first block would be compared against a window the
+        // previous endpoint validated, which is exactly the spurious-rollback
+        // shape the suspension exists to prevent.
+        if switch_watcher.switched(cached_provider.as_ref()) {
```
hane-polymarket, line 1283: **[P1] Detect failover during `get_logs`**

The switch counter is checked after `get_latest_block`, but `get_logs` can itself fail over later in the iteration. Its range was calculated from the old endpoint's tip, so a lagging replacement can successfully return an empty result for blocks it has not reached; the loop then advances past those blocks permanently. Check the counter immediately after `get_logs` and discard/retry the range when it changed, or verify the replacement tip covers `to_block`.

core/src/provider.rs
```diff
@@ -1158,57 +1282,137 @@ pub async fn create_client(
...
+                );
+                continue;
+            }
+
+            let parsed_url = Url::parse(rpc_url).map_err(|e| {
+                RetryClientError::ProviderCantBeCreated(rpc_url.to_string(), e.to_string())
+            })?;
+
+            let health = Arc::new(EndpointHealth::new(rpc_url));
+            let transport = RpcLoggingLayer::new(chain_id, rpc_url.to_string())
```
hane-polymarket, line 1355: **[P2] Redact fallback credentials in RPC logs**

This passes every fallback's raw configured URL into `RpcLoggingLayer`, whose slow/error paths log that URL and sometimes the raw transport error. Since the background prober continuously calls every fallback, credentials embedded in paths or query parameters can leak even when an endpoint never serves indexing traffic. Pass the redacted URL and redact errors before logging.

hane-polymarket (commented):
I found three correctness/security issues in the new failover path.

PR #103, opened 2024-09-16
Hypersync source support

SozinM:
This Pr adds hypersync integration to the rindexer.
Initially i made it with enum, but i think that dynamic dispatch fits better here. We need to choose instance during runtime via config.

This is p1 of the changes, would be nice to move streaming logic to another trait/extend current one, to be able to specialize streaming based on the upstream.

PR #457, opened 2026-07-24
feat: HyperSync support as alternative log data source

moose-code:
Revives #103 — HyperSync as an alternative log data source — rebuilt against the modern `ChainProvider` trait.

## Why the #103 blocker is gone

#103 was closed because `hypersync-net-types 0.9` required the `capnp` compiler installed on every build machine. Since 0.12 the crate ships pre-generated code and has **no `build.rs`** — verified: a clean `cargo build` needs nothing beyond cargo.

## Design

`HypersyncProvider` implements the existing `ChainProvider` trait and wraps the network's `JsonRpcCachedProvider`:

- **Historical `get_logs` → HyperSync.** Everything else — `eth_call`, receipts, traces, latest-block polling — stays on the `rpc` endpoint, which remains required.
- **Near-tip requests → RPC.** A cached, monotonic archive-height check routes any request past HyperSync's ingested height to the RPC node, so live indexing, bloom checks, and reorg handling are completely unchanged. If HyperSync is down or lagging, everything silently degrades to today's RPC behavior (its errors also flow into the existing retry-with-halving path).
- **Block timestamps are joined** into each HyperSync query and stamped on logs, so with `timestamps: true` the block clock skips its `eth_getBlockByNumber` batching entirely (stage 1 of its waterfall).
- Everything above `get_logs` — factory contracts, parallel backfill, reorg windows, dependency ordering — is untouched.

Rust-project codegen emits a hypersync-wrapped indexing provider too, and `read_manifest` un-substitutes `hypersync.url` / `hypersync.api_token` for rust projects the same way it does for `rpc`, so tokens never land in generated code. The user-facing `get_X_provider()` helpers remain plain RPC.

## Config

```yaml
networks:
- name: ethereum
  chain_id: 1
  rpc: https://mainnet.gateway.tenderly.co
  hypersync: true   # shorthand: https://{chain_id}.hypersync.xyz + token from env
```

Full form: `hypersync: {url, api_token, max_block_range}`. Token falls back to `HYPERSYNC_API_TOKEN` / `ENVIO_API_TOKEN` (free tier at envio.dev/app/api-tokens). `max_block_range` is the window per HyperSync request — default 50,000 keeps dense contracts memory-safe; sparse contracts benefit from much wider windows (the full-history run below used 2M).

## Benchmarks

Same release binary, same manifest per row — only the data source differs. Timings are rindexer's own `Historical indexing completed` measure. Every completed run was verified: identical row counts everywhere, and byte-identical sorted output (MD5) spot-proven on the 8.3M-row run.

**One day of Polymarket pUSD, Polygon, 8,338,067 Transfer events over 43,200 blocks (CSV):**

| Source | Time | vs HyperSync |
|---|---|---|
| HyperSync | **3m 22s** | — |
| dRPC (paid) | 16m 48s | 5.0× |
| Tenderly (free) | 21m 1s | 6.2× |
| Dwellir (paid) | stopped at 10 min / 3.97% — projected ≈ 4.2h | ≈ 75× |

**The rest of the matrix (Ethereum mainnet):**

| Workload | Events | HyperSync | Tenderly (free) | dRPC (paid) |
|---|---|---|---|---|
| USDC, 100k blocks (dense) | 938,890 | **10s** | 3m 49s · 23× | 17m 56s · 107× |
| USDC + `timestamps: true` → Postgres, 10k recent blocks | 925,103 | **17s** | 1m 8s · 4.0× | 11m 49s · 42× |
| rETH, first 2M blocks (sparse, 2M window) | 23,207 | **12s** | 22s · 1.8× | 2m 32s · 12.7× |
| rETH, entire history block 0 → 25.6M | 901,220 | **2m 53s** | stopped at 10 min with 6% of events | — |
| One week of pUSD (HyperSync showcase) | 57,385,711 | **16m 13s** | — | — |

Notes on reading it honestly:

- **Provider variance is the real story.** Tenderly free (the provider the rindexer docs recommend as fastest) is consistently the best RPC because it predicts block ranges; dRPC's paid plan trails it 3–6× due to response/range caps. HyperSync costs the same regardless of provider quirks.
- **The sparse row is the fair caveat**: with a good range-predicting provider, a bounded sparse window closes to 1.8×. The big wins are dense data and whole-chain scans.
- **Timestamps demo**: that window carried logs in 9,994 distinct blocks past the shipped block-clock data — the RPC runs resolved all 9,994 timestamps over the wire, HyperSync resolved zero, and all 925,103 `(tx_hash, log_index, block_timestamp)` rows are identical across all three Postgres databases.
- **Request efficiency** (measured through a counting proxy): backfill averages ~10,300 events per HyperSync request (~370 KB responses).

## Testing

- 734 lib tests pass (new coverage: manifest parsing incl. the `hypersync: true` shorthand, codegen output, secrets un-substitution); clippy and fmt clean.
- Live indexing validated at the chain tip on USDC: block-by-block real time, zero errors, near-tip requests observed falling back to RPC as designed. The provider records `hypersync_getLogs` vs `eth_getLogs` separately in `rindexer_rpc_requests_total`, so the backend split is visible on `/metrics`.

## Follow-ups (not in this PR)

- Density-adaptive HyperSync window sizing (auto-widen on sparse contracts instead of the `max_block_range` knob).
- Use HyperSync's bloom `address_filter` for factory contracts with very large child sets.
- An e2e test behind an env-gated API token.

### DIFF
--- a/core/src/hypersync.rs
+++ b/core/src/hypersync.rs
@@ -362,7 +362,21 @@ impl ChainProvider for HypersyncProvider {
             start.elapsed().as_secs_f64(),
         );
 
-        result
+        match result {
+            Ok(logs) => Ok(logs),
+            Err(e) => {
+                // A HyperSync failure degrades to RPC rather than failing the fetch.
+                // This also covers a load-balanced instance lagging the height we
+                // checked: the client errors loudly on zero progress instead of
+                // stalling, and the range is served by RPC. If the range is too wide
+                // for the RPC node, its error feeds the fetch loop's usual adaptive
+                // range-halving.
+                warn!(
+                    "HyperSync get_logs failed for blocks [{from_block}..{to_block}], falling back to RPC: {e:#}",
+                );
+                self.rpc.get_logs(event_filter).await
+            }
+        }
     }
 
     // Blocks, receipts and traces could also be served from HyperSync, but faithfully
@@ -422,5 +436,69 @@ impl ChainProvider for HypersyncProvider {
 
     async fn eth_call_latest(&self, to: Address, data: Bytes) -> Result<String, ProviderError> {
         self.rpc.eth_call_latest(to, data).await
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use alloy::transports::mock::Asserter;
+
+    use super::*;
+
+    /// A provider whose HyperSync endpoint refuses connections (nothing listens on the
+    /// target port) and whose RPC side is a mock that serves pushed responses.
+    fn provider_with_unreachable_hypersync() -> (HypersyncProvider, Asserter) {
+        let (rpc, asserter) = JsonRpcCachedProvider::mock_with_asserter(1);
+        let client = Client::builder()
+            .url("http://127.0.0.1:9")
+            .api_token("00000000-0000-0000-0000-000000000000")
+            .max_num_retries(0)
+            .build()
+            .expect("building a client makes no network calls");
+
+        let provider = HypersyncProvider {
+            client,
+            rpc,
+            max_block_range: None,
+            stream_config: StreamConfig::default(),
+            height_cache: Mutex::new(None),
+        };
+
+        (provider, asserter)
+    }
+
+    #[tokio::test]
+    async fn get_logs_falls_back_to_rpc_when_height_check_fails() {
+        let (provider, asserter) = provider_with_unreachable_hypersync();
+        asserter.push_success(&Vec::<Log>::new());
+
+        let logs = provider.get_logs(&RindexerEventFilter::empty_for_test()).await.unwrap();
+
+        assert!(logs.is_empty(), "expected the mocked RPC response, got {logs:?}");
+    }
+
+    #[tokio::test]
+    async fn get_logs_falls_back_to_rpc_when_hypersync_query_errors() {
+        let (provider, asserter) = provider_with_unreachable_hypersync();
+        // Seed the height cache so `covers_block` passes and the HyperSync query path
+        // runs — and errors against the unreachable endpoint — instead of the request
+        // being routed to RPC by the height gate.
+        *provider.height_cache.lock().await = Some((Instant::now(), u64::MAX));
+        asserter.push_success(&Vec::<Log>::new());
+
+        let logs = provider.get_logs(&RindexerEventFilter::empty_for_test()).await.unwrap();
+
+        assert!(logs.is_empty(), "expected the mocked RPC response, got {logs:?}");
+    }
+
+    #[tokio::test]
+    async fn covers_block_trusts_cached_height_at_or_past_target() {
+        let (provider, _asserter) = provider_with_unreachable_hypersync();
+        *provider.height_cache.lock().await = Some((Instant::now(), 100));
+
+        // At or below the cached height: trusted without a network call.
+        assert!(provider.covers_block(100).await);
+        // Past the cached height within the TTL: not covered, no refresh attempted.
+        assert!(!provider.covers_block(101).await);
     }
 }
--- a/core/src/hypersync.rs
+++ b/core/src/hypersync.rs
@@ -54,7 +54,7 @@ pub struct HypersyncProvider {
     /// Fallback provider used for everything HyperSync cannot serve.
     rpc: Arc<JsonRpcCachedProvider>,
     max_block_range: Option<U64>,
-    stream_concurrency: Option<usize>,
+    stream_config: StreamConfig,
     height_cache: Mutex<Option<(Instant, u64)>>,
 }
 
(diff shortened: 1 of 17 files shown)


DZakh:
Hi, Envio engineers here 👋 We are aware of the architectural, logical, and performance issues of the Draft, and we are ready to polish the PR further until it satisfies our and your quality requirements. I believe HyperSync as a source will benefit us in terms of new potential users, as well as you having a performant source for backfill - since Rindexer is written in Rust and has performance and data quality as main priorities, I hope this is a win-win for both of us 👍

Let us know what you think 🙏

joshstevens19:
Hey happily work with you all to get this merged big fans of what you guys build. Away now till next weekend but why don’t you get the PR in the ready state I can review then and merge and release - excited to see it!

core/src/hypersync.rs
```diff
@@ -0,0 +1,413 @@
...
+                LogField::Topic3,
+            ])
+            .select_block_fields([BlockField::Number, BlockField::Timestamp]);
+
+        // `collect_arrow` paginates internally until the full requested range is covered,
+        // so a successful return always means complete coverage of [from_block, to_block].
+        // The arrow response is materialized straight into alloy types via the arrow row
+        // readers, skipping the intermediate simple-types allocation pass.
+        let response =
+            self.client.collect_arrow(query, self.stream_config.clone()).await.map_err(map_err)?;
```
JonoPrest, line 242: The one thing I'm unsure of here is the following scenario:

Block N (say 100) is available on one of our hypersync instances but not all yet. Maybe an external rpc has up to 101 for argument sake.

If Rindexer asks for logs up to block 100 or maybe even 101 if the fallback rpc has this number and then we hit an instance of hypersync that only has up to block 99. I'm not sure if the collect query handles this scenario - essentially it should do a backoff and retry until the block is available. This is what hyperindex does.

core/src/hypersync.rs
```diff
@@ -0,0 +1,426 @@
...
+
+        // The archive lags the chain head by a few blocks, so near-tip requests (live
+        // indexing) go to the RPC node, which is authoritative for the head.
+        if !self.covers_block(to_block).await {
+            return self.rpc.get_logs(event_filter).await;
+        }
+
+        let start = Instant::now();
+        let result =
+            self.get_logs_via_hypersync(event_filter, addresses, from_block, to_block).await;
```
joshstevens19, line 356: Can we fallback to `self.rpc.get_logs(event_filter).await` when the HyperSync query itself errors? Right now we only fallback if `/height` fails or lags. If `/height` succeeds and then `collect_arrow` hits a 5xx/timeout/etc, the fetch loop keeps shrinking and retrying HyperSync, so a healthy RPC cannot save the run. That also contradicts the PR/docs claim that HyperSync being unavailable degrades to RPC.

core/src/manifest/network.rs
```diff
@@ -70,6 +165,18 @@ pub struct Network {
...
+    /// The `rpc` endpoint is still required for everything HyperSync cannot serve.
+    ///
+    /// Accepts `hypersync: true` as shorthand for the default endpoint
+    /// (`https://{chain_id}.hypersync.xyz`) or a config map for customisation.
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        deserialize_with = "deserialize_option_hypersync"
+    )]
+    pub hypersync: Option<HypersyncConfig>,
```
joshstevens19, line 178: Need to wire this into `core/src/hot_reload/diff.rs`. `diff_networks` currently only compares `rpc`, `chain_id`, `block_poll_frequency`, `compute_units_per_second`, and `max_block_range`, so changing `hypersync` while `rindexer start --watch` is running can be treated as `NoChange` and keep using the old provider. I’d add the serialized `hypersync` config to the network diff and cover it with a focused hot-reload test.

documentation/docs/pages/docs/start-building/yaml-config/networks.mdx
```diff
@@ -100,6 +100,57 @@ networks:
   max_block_range: 10000 // [!code focus]
 ```
 
+### hypersync
```
joshstevens19, line 103: Need an unreleased changelog entry before merge. This is a new user-facing YAML option/dependency/runtime path, so per the repo rules this wants a `- feat:` bullet at the top of `documentation/docs/pages/docs/changelog.mdx` naming `networks[].hypersync` and the token/default behaviour.

joshstevens19 (changes requested):
AI: Thanks for the work here, big fan of this direction and the backfill numbers look great. I think this is close but there are a few bits we need to tighten before merging. I checked the PR locally too: `cargo check -p rindexer --all-targets` passed, and the focused HyperSync manifest/codegen nextest run passed (`5 tests`).

core/src/hypersync.rs
moose-code: Done, `get_logs` now degrades to the wrapped RPC provider on any HyperSync error, with unit tests covering both the height-gate and query-error paths against a mocked RPC. Oversized fallback ranges shrink through the existing adaptive halving, same as RPC-only runs.

core/src/manifest/network.rs
moose-code: done 👍

documentation/docs/pages/docs/start-building/yaml-config/networks.mdx
moose-code: Added 👍

joshstevens19 (approved):

PR #458, opened 2026-08-03
fix(reorg): support iterated custom table rollbacks

SkandaBhat:
## Summary

- retain custom-table `iterate` bindings in derived rollback plans
- expand paired arrays and compile manifest event conditions when snapshotting orphaned deltas
- serialize ClickHouse rollback join keys so composite/`UInt256` custom-table keys work

## Why

Rollback snapshots treated iterate aliases as physical event-table columns and interpolated conditions such as `$from` and `$to` as raw SQL. ERC-1155 `TransferBatch` rollback therefore failed before orphaned rows could be removed. ClickHouse `StorageJoin` also rejected custody-style composite/`UInt256` keys.

## Validation

- exact YAML-to-rollback-plan regression
- PostgreSQL + ClickHouse `TransferBatch` reorg integration covering paired `ids`/`values`, both add/subtract inverses, raw-row deletion, and checkpoint rewind
- default-feature workspace suite: 814 passed
- all-features workspace suite: 826 passed
- `cargo clippy --workspace --all-targets --locked -- -D warnings`
- `cargo fmt --all -- --check`

### DIFF
--- a/core/src/event/filter/ast.rs
+++ b/core/src/event/filter/ast.rs
@@ -214,14 +214,14 @@ impl<'a> Expression<'a> {
         })
     }
 
-    pub(crate) fn to_sql_event_condition<F>(&self, event_column: F) -> Option<String>
+    pub(crate) fn to_sql_event_condition<F>(&self, event_variable: F) -> Option<String>
     where
-        F: Fn(&str) -> String,
+        F: Fn(&ConditionLeft<'_>) -> String,
     {
         if self.has_table_references() {
             return None;
         }
-        Some(self.to_sql_condition_with(&|variable| event_column(variable.base_name())))
+        Some(self.to_sql_condition_with(&event_variable))
     }
 
     fn to_sql_condition_with<F>(&self, variable_to_sql: &F) -> String
--- a/core/src/event/filter/ast.rs
+++ b/core/src/event/filter/ast.rs
@@ -205,10 +205,33 @@ impl<'a> Expression<'a> {
     /// - Event variables ($var) become EXCLUDED.var (the new values being inserted)
     /// - Table variables (@var) become {table_name}.var (current DB values)
     pub fn to_sql_condition(&self, table_name: &str) -> String {
+        self.to_sql_condition_with(&|variable| {
+            let column = variable.base_name();
+            match variable.source() {
+                VariableSource::Event => format!("EXCLUDED.\"{}\"", column),
+                VariableSource::Table => format!("{}.\"{}\"", table_name, column),
+            }
+        })
+    }
+
+    pub(crate) fn to_sql_event_condition<F>(&self, event_column: F) -> Option<String>
+    where
+        F: Fn(&str) -> String,
+    {
+        if self.has_table_references() {
+            return None;
+        }
+        Some(self.to_sql_condition_with(&|variable| event_column(variable.base_name())))
+    }
+
+    fn to_sql_condition_with<F>(&self, variable_to_sql: &F) -> String
+    where
+        F: Fn(&ConditionLeft<'_>) -> String,
+    {
         match self {
             Expression::Condition(cond) => {
-                let left_sql = cond.left.to_sql(table_name);
-                let right_sql = cond.right.to_sql(table_name);
+                let left_sql = cond.left.to_sql_with(variable_to_sql);
+                let right_sql = cond.right.to_sql_with(variable_to_sql);
                 let op_sql = match cond.operator {
                     ComparisonOperator::Eq => "=",
                     ComparisonOperator::Ne => "<>",
@@ -220,8 +243,8 @@ impl<'a> Expression<'a> {
                 format!("{} {} {}", left_sql, op_sql, right_sql)
             }
             Expression::Logical { left, operator, right } => {
-                let left_sql = left.to_sql_condition(table_name);
-                let right_sql = right.to_sql_condition(table_name);
+                let left_sql = left.to_sql_condition_with(variable_to_sql);
+                let right_sql = right.to_sql_condition_with(variable_to_sql);
                 let op_sql = match operator {
                     LogicalOperator::And => "AND",
                     LogicalOperator::Or => "OR",
@@ -229,7 +252,7 @@ impl<'a> Expression<'a> {
                 format!("({} {} {})", left_sql, op_sql, right_sql)
             }
             Expression::Not(inner) => {
-                format!("NOT ({})", inner.to_sql_condition(table_name))
+                format!("NOT ({})", inner.to_sql_condition_with(variable_to_sql))
             }
         }
     }
@@ -250,22 +273,29 @@ impl<'a> ArithmeticExpr<'a> {
     /// Converts this arithmetic expression to SQL.
     /// Column names are double-quoted to handle reserved keywords safely.
     pub fn to_sql(&self, table_name: &str) -> String {
-        match self {
-            ArithmeticExpr::Variable(cond_left) => {
-                let col_name = cond_left.base_name();
-                match cond_left.source() {
-                    VariableSource::Event => format!("EXCLUDED.\"{}\"", col_name),
-                    VariableSource::Table => format!("{}.\"{}\"", table_name, col_name),
-                }
+        self.to_sql_with(&|variable| {
+            let column = variable.base_name();
+            match variable.source() {
+                VariableSource::Event => format!("EXCLUDED.\"{}\"", column),
+                VariableSource::Table => format!("{}.\"{}\"", table_name, column),
             }
+        })
+    }
+
+    fn to_sql_with<F>(&self, variable_to_sql: &F) -> String
+    where
+        F: Fn(&ConditionLeft<'_>) -> String,
+    {
+        match self {
+            ArithmeticExpr::Variable(variable) => variable_to_sql(variable),
             ArithmeticExpr::Literal(lit) => match lit {
                 LiteralValue::Bool(b) => if *b { "TRUE" } else { "FALSE" }.to_string(),
                 LiteralValue::Str(s) => format!("'{}'", s.replace('\'', "''")),
                 LiteralValue::Number(n) => n.to_string(),
             },
             ArithmeticExpr::Binary { left, operator, right } => {
-                let left_sql = left.to_sql(table_name);
-                let right_sql = right.to_sql(table_name);
+                let left_sql = left.to_sql_with(variable_to_sql);
+                let right_sql = right.to_sql_with(variable_to_sql);
                 match operator {
                     ArithmeticOperator::Power => {
                         // Use PostgreSQL POWER() function for exponentiation
--- a/core/src/indexer/reorg/task.rs
+++ b/core/src/indexer/reorg/task.rs
@@ -1,7 +1,4 @@
-use std::{
-    collections::{HashMap, HashSet},
-    sync::Arc,
-};
+use std::{collections::HashMap, sync::Arc};
 
 use alloy::primitives::{B256, U64};
 use anyhow::Context;
@@ -99,7 +96,7 @@ impl DerivedColumnRollback {
         action: SetAction,
     ) -> anyhow::Result<Self> {
         super::validate_sql_identifier(&derived_column, "derived column")?;
-        validate_event_operand(&event_column, "rollback event column")?;
+        parse_event_operand(&event_column, "rollback event column")?;
         Ok(Self { derived_column, event_column, action })
     }
 }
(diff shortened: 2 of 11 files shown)


ExoMonk (approved):

core/src/indexer/reorg/task.rs
```diff
@@ -345,26 +451,64 @@ impl ReorgTask {
...
+                    }
+                    let group_select = group_select.join(", ");
                     let aggs = agg_specs
                         .iter()
                         .map(|(is_counter, ev, alias)| {
                             if *is_counter {
                                 format!("COUNT(*) AS {}", alias)
                             } else {
-                                format!("SUM({}) AS {}", quote(ev), alias)
+                                format!("SUM({}) AS {}", source.column(ev), alias)
```
hane-polymarket: [P1] Preserve numeric types for iterated PostgreSQL values

The generated PostgreSQL event schema maps ABI `uint256[]` to `VARCHAR(78)[]`, so `unnest(values)` produces `varchar`. `SUM(source.column)` therefore fails with `function sum(character varying) does not exist`; the later `NUMERIC = VARCHAR` key join would fail too. The new E2E fixture declares `ids` and `values` as `NUMERIC[]`, masking the actual ERC-1155 schema. Please carry source and derived types into the rollback plan and cast both aggregates and snapshot keys, then test with the generated ABI types.

core/src/event/filter/ast.rs
```diff
@@ -205,10 +205,33 @@ impl<'a> Expression<'a> {
...
+    }
+
+    pub(crate) fn to_sql_event_condition<F>(&self, event_column: F) -> Option<String>
+    where
+        F: Fn(&str) -> String,
+    {
+        if self.has_table_references() {
+            return None;
+        }
+        Some(self.to_sql_condition_with(&|variable| event_column(variable.base_name())))
```
hane-polymarket: [P2] Preserve accessors when compiling event conditions

Passing only `base_name()` drops valid paths. `$data.amount > 0` becomes a reference to raw column `data` instead of flattened `data_amount`; `$ids[0]` becomes the whole `ids` array. Either case can make snapshot creation fail or evaluate differently from forward processing. Pass the full `ConditionLeft` to the renderer and handle key and index accessors per dialect, with nested and indexed regression cases.

hane-polymarket (commented):
Two net-new correctness issues found on the current head.

core/src/indexer/reorg/task.rs
skanda-polymarket: Fixed in 4fd1dc97. Rollback metadata now carries ABI source types and resolved derived-column types. PostgreSQL casts snapshot keys and aggregate inputs before grouping and SUM, and the TransferBatch regression uses generated VARCHAR(78)[] raw-event columns.

core/src/event/filter/ast.rs
skanda-polymarket: Fixed in 4fd1dc97. Event-condition SQL rendering now receives the full ConditionLeft and preserves flattened key access plus 0-based manifest array access through dialect-specific SQL. Added PostgreSQL and ClickHouse regressions for $data.amount and $ids[0].

core/src/indexer/start.rs
```diff
@@ -510,12 +589,19 @@ fn build_derived_tables_for_event(
...
-                        operation.condition().map(String::from),
-                    )?);
+                    rollback_ops.push(
+                        DerivedTableRollbackOp::try_new(
+                            event_table_name.clone(),
+                            where_columns,
+                            columns,
+                            operation.condition().map(String::from),
+                        )?
+                        .with_iterate(table_event.iterate.clone())?
```
hane-polymarket: [P1] Preserve accessors in rollback operands

The `where_columns` and reversible `columns` passed into this newly extended rollback plan are still collapsed with `camel_to_snake`, so supported expressions such as `$data.amount` and `$ids[0]` become `dataamount` and `ids0` rather than the raw event columns/accessors used by the forward path. Reorg snapshot creation will therefore reference a nonexistent column or replay the wrong expression for nested/indexed updates and upserts. Please retain the parsed source expression/accessor path in the rollback metadata and render it through the same dialect-aware accessor logic used for conditions.

core/src/indexer/reorg/task.rs
```diff
@@ -106,6 +108,12 @@ pub struct DerivedTableRollbackOp {
     pub columns: Vec<DerivedColumnRollback>,
     /// Optional SQL condition re-evaluated against event data.
     pub condition: Option<String>,
+    /// Array bindings expanded by the forward custom-table operation.
+    pub iterate: Vec<IterateBinding>,
+    /// Semantic types of source event columns and iterate aliases.
+    pub source_column_types: HashMap<String, ColumnType>,
```
hane-polymarket: [P2] Account for the public struct-literal break

`DerivedTableRollbackOp` is publicly exported, and adding `iterate`, `source_column_types`, and `derived_column_types` as required public fields breaks downstream struct literals at compile time (the integration-test literals in this PR already had to be updated). The changelog's Breaking changes section is still empty. Please either keep this metadata behind a non-breaking internal representation/preserve the public shape, or explicitly classify the change as breaking and provide the required migration guidance.

hane-polymarket (commented):
Two net-new issues remain after re-reviewing the latest head.

core/src/indexer/start.rs
skanda-polymarket: Fixed in dfd91f01. Rollback WHERE and SET operands now retain the full event accessor path and are rendered through the same dialect-aware accessor logic as conditions. Added regressions for nested tuple fields and indexed arrays, plus fail-closed handling for unreconstructable table-state conditions.

core/src/indexer/reorg/task.rs
skanda-polymarket: Fixed in dfd91f01 without changing the existing DerivedTableRollbackOp shape. Iterate/type metadata now lives in an optional DerivedTableRollbackPlan sidecar; existing struct literals and ReorgTask::execute callers remain source-compatible, while the runtime uses the additive execute_with_rollback_plan path.

core/src/indexer/reorg/task.rs
```diff
@@ -345,39 +743,99 @@ impl ReorgTask {
...
+                    ))
                 };
 
                 if let Some(pg) = pg {
                     let pg_temp = format!("{}_pg", temp_base);
                     let pg_create = format!(
                         "CREATE TEMP TABLE {} AS {}",
                         pg_temp,
-                        build_select(&quote_pg_ident)
+                        build_select(SqlDialect::Postgres)?
```
hane-polymarket, line 1040: [P1] Keep the PostgreSQL snapshot on one connection

This temporary table is created through `PostgresClient::batch_execute`, which checks out a pooled connection and releases it when the call returns. `reorg_rollback_transaction` and the later snapshot `UPDATE` each check out independently, so under concurrent pool use the update can run in a different session where this table does not exist. By then raw-event deletion has committed, leaving derived state unreversed. Please pin one PostgreSQL connection—preferably one transaction—across snapshot creation, raw rollback, reversal, and cleanup.

core/src/indexer/reorg/task.rs
```diff
@@ -250,6 +341,235 @@ fn quote_ch_ident(name: &str) -> String {
...
+        match self.dialect {
+            SqlDialect::Postgres => {
+                let aliases = self
+                    .iterate
+                    .iter()
+                    .map(|binding| self.dialect.quote(&binding.alias))
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                format!(
+                    "{} AS _rindexer_event CROSS JOIN LATERAL unnest({}) AS _rindexer_iter({})",
```
hane-polymarket: [P2] Expand PostgreSQL tuple arrays as JSONB

Generated PostgreSQL event schemas store `tuple[]` and `tuple[N]` fields as `JSONB`, while this path always emits native-array `unnest` and the preceding length check always emits `cardinality`. Forward `expand_iterate_bindings` accepts tuple arrays, so a valid configuration can index successfully but every reorg aborts with a PostgreSQL type error. Please carry the storage representation into the rollback plan and use `jsonb_array_length`/`jsonb_array_elements` plus tuple-field extraction for these bindings.

core/src/indexer/reorg/task.rs
```diff
@@ -404,13 +862,10 @@ impl ReorgTask {
                     // aggregates via joinGet(). ClickHouse mutations don't
                     // support correlated subqueries the way Postgres does, so
                     // we cannot use `(SELECT ... WHERE snap.k = dt.k LIMIT 1)`.
-                    let ch_snap_keys: Vec<String> =
-                        where_ev_cols.iter().map(|ev_col| quote_ch_ident(ev_col)).collect();
                     let ch_create = format!(
-                        "CREATE TABLE IF NOT EXISTS {} ENGINE = Join(ANY, LEFT, {}) AS {}",
+                        "CREATE TABLE IF NOT EXISTS {} ENGINE = Join(ANY, LEFT, _rindexer_key) AS {}",
```
hane-polymarket, line 1068: [P2] Do not reuse stale ClickHouse snapshots

The snapshot name is deterministic from network, fork point, and operation index, and `IF NOT EXISTS` silently keeps any table left by an interrupted run. A later reorg that resolves to the same name then applies obsolete aggregates instead of the new `SELECT`. Please make the name execution-scoped with a unique nonce or explicitly drop and recreate it, with cleanup on every failure path.

hane-polymarket (commented):
Three net-new correctness issues found on the current head.

core/src/indexer/reorg/task.rs
skanda-polymarket: Fixed in 62984519. PostgreSQL snapshot creation, raw deletion and checkpoint rewind, derived reversal, journal recalculation, and insert-only cleanup now share one pinned transaction. The E2E forces reversal SQL to fail after deletion and verifies the raw row and checkpoint both roll back.

core/src/indexer/reorg/task.rs
skanda-polymarket: Fixed in 62984519. Rollback metadata now distinguishes native arrays from PostgreSQL JSONB tuple arrays, using jsonb_array_length, jsonb_array_elements, and typed JSON-path extraction. The E2E covers a camelCase tuple field zipped with a native array.

core/src/indexer/reorg/task.rs
skanda-polymarket: Fixed in 62984519. ClickHouse snapshot names now include a 24-character execution nonce, CREATE no longer reuses existing tables, and centralized reverse-order cleanup runs after success and every captured failure. The E2E plants a stale legacy snapshot and verifies it is ignored with no new snapshot leak.

hane-polymarket (commented):
Two net-new correctness issues found on the current head.

core/src/indexer/reorg/task.rs
```diff
@@ -830,20 +1450,39 @@ impl ReorgTask {
...
+        };
+        let pg = pg_transaction.as_ref();
 
-        let mut affected_tx_hashes: Vec<String> = Vec::new();
-        let mut total_deleted = 0u64;
+        // One nonce scopes every snapshot created by this execution. PostgreSQL
+        // still relies on its session-local temp namespace; ClickHouse needs the
+        // nonce because its snapshot tables are database-global.
+        let attempt_id = generate_random_id(24).to_ascii_lowercase();
+        let reversal_snapshots =
```
hane-polymarket: [P1] Quiesce event writers before taking rollback snapshots

The coordinator mutex serializes detection paths, but the buffered fetch pipeline can already have spawned callbacks, and those callbacks never acquire this mutex. In the PostgreSQL sole-sink path, custom-table operations deliberately commit before the raw-event insert; a callback can therefore commit the derived delta, this snapshot can miss its raw row, and the later READ COMMITTED delete can remove that row after it arrives without any corresponding reversal. Please drain or block all in-flight writers for this network before snapshotting and keep that barrier through rollback.

core/src/indexer/reorg/task.rs
```diff
@@ -853,96 +1492,103 @@ impl ReorgTask {
...
-                );
+                    );
+                }
             }
+
+            Ok((total_deleted, affected_tx_hashes))
+        }
+        .await;
+
+        Self::cleanup_reversal_snapshots(&reversal_snapshots, pg, clickhouse).await;
```
hane-polymarket: [P1] Preserve ClickHouse reversal data across partial failures

By this point `ClickhouseClient::reorg_rollback` has synchronously deleted the raw event rows, but `apply_reversal_from_snapshots`, journal recalculation, or a later derived-table mutation may still have failed. This unconditional cleanup then drops the only copy of the orphaned deltas and returns the error. Because the coordinator leaves its window unchanged, the next attempt retries, snapshots an already-empty ClickHouse event table, and can never reverse the derived state. Please retain durable retry state/snapshots until every ClickHouse phase succeeds, or reorder and checkpoint the phases so a retry can resume idempotently.

hane-polymarket (approved):

joshstevens19 (approved):

PR #460, opened 2026-08-07
fix: arithmetic upserts in no-code tables losing debits on row creation

joshstevens19:
## Problem

The [envio open-indexer-benchmark `erc20-account-balances` case](https://github.com/enviodev/open-indexer-benchmark/blob/main/cases/erc20-account-balances/README.md#rindexer) reported that rindexer's no-code tables could not compute running balances: debits were intermittently lost (465 of 1,747 accounts absent, 672 with wrong balances), and reordering the YAML operations only changed *which* addresses broke.

**Root cause:** arithmetic (`add`/`subtract`) upserts compile to one `INSERT ... ON CONFLICT DO UPDATE`, but the arithmetic only existed on the conflict branch:

```sql
INSERT INTO balances (...) SELECT tp.network, tp.holder, tp.balance ...   -- raw +value
ON CONFLICT (network, holder)
DO UPDATE SET balance = COALESCE(balances.balance, 0) - EXCLUDED.balance
```

A row *created* by a `subtract` — any address whose first touch inside the indexed window is a debit — got `+value` instead of `default − value`. Which addresses hit that depends on batch boundaries and operation order, hence "intermittent". The bug ships in **every release since the tables feature landed (v0.30.0 → v0.41.0)** and affects the flagship docs example (`tables_erc20_balances`) plus `tables_erc1155_balances`, `tables_dex_pool`, and any subtract-bearing table indexed from mid-history. Add-only and set-only tables were unaffected, and indexing from a token's deployment block masks it (everyone receives before sending) — which is why it survived so long.

## Fix

- The INSERT branch now selects `(default ± delta)`; the conflict branch recovers the delta via `col = COALESCE(t.col, 0) + EXCLUDED.col - default`. One clause serves add and subtract — the sign lives in the inserted value.
- `EXCLUDED."col"` references to arithmetic columns inside SQL-pushed `if:` conditions (e.g. `"$value <= @balance"`) are rewritten back to the raw delta, since `EXCLUDED` now carries `default ± delta` — without this, subtract guards would silently invert.
- YAML column defaults are canonicalized through strict numeric parsing (`u64`/`i64`/`U256`/`I256`) before touching SQL text; unparseable defaults (e.g. `"1e18"`, which Postgres DDL accepts but would diverge) fall back to 0 with a startup warning.

### Custom Rust indexing (`create_batch_postgres_operation!`) brought to parity

The macro had the same sign bug **plus** two bugs #389 only fixed for the no-code path: `DISTINCT ON` silently dropped duplicate-key deltas within a batch, and the sequence guard silently dropped arithmetic updates from lower-sequence events. The macro now uses the same GROUP BY aggregation (gated off for plain `Insert` ops), guard-disable for arithmetic, latest-by-sequence set columns, and gains `max`/`min` support. Array-typed set columns aggregate via `MAX` (`array_agg(...)[1]` yields NULL for arrays; `MAX(bool/bytea)` doesn't exist, so scalars use `array_agg`). Note: `rindexer::database` is private, so the exported macro currently can't compile in external crates anyway (E0603) — worth a follow-up decision.

### Bonus fix: table-only projects couldn't index at all since v0.37.0

Found while validating end-to-end: since #375, the runtime stores raw events for table-driving events (they're the reorg recomputation source of truth), but the DDL only created raw event tables for `include_events` — so a no-code project with `tables:` and no `include_events` COPYs into a missing relation and retries forever. Nothing indexes; the docs' flagship example is unrunnable on current master. DDL (Postgres + ClickHouse), migrations, and the runtime now share a single `Contract::stores_raw_events()` predicate.

## Verification

- **Mutation check:** reverting just the SQL fix makes the new integration test fail with the benchmark's exact corruption signature (sender-first account `+9` instead of `−5`).
- **Real end-to-end run:** `rindexer start indexer` over mainnet USDC blocks 24250761–24250860 (6,675 Transfer events) compared against balances folded independently from raw `eth_getLogs`: **4,057 accounts — 0 missing, 0 extra, 0 wrong**, including all 1,373 negative-balance accounts (the class this bug corrupted).
- **New tests** (testcontainers, run per-process under nextest like the existing docker tests): benchmark-shape running-balances fold, non-zero column defaults, sequence-guard regression (stale write applied last), guarded-condition semantics with zero and non-zero defaults, max upserts (smaller value applied last), macro parity (duplicate-key accumulation, lower-sequence deltas, latest-by-sequence set columns, max columns), pure-SQL unit tests for every new builder, and a DDL test for table-only raw event tables.
- Full suite: 806/806 passing, `cargo clippy --workspace --all-targets -- -D warnings` clean, fmt clean.

## Known pre-existing issues (out of scope, worth separate issues)

1. Shutdown mid-batch can checkpoint a block for which only some operations ran — the remaining operations are skipped on restart.
2. A mid-operations error triggers infinite callback retry, re-applying already-committed arithmetic statements (arithmetic upserts are at-least-once by design now; documented on the macro).
3. ClickHouse custom tables never accumulate `add`/`subtract` (plain INSERT into `ReplacingMergeTree` keeps only the last event's value) — needs a separate design.
4. The exported macro is unusable from external crates due to `mod database;` being private.

### DIFF
--- a/core/src/database/batch_operations.rs
+++ b/core/src/database/batch_operations.rs
@@ -82,6 +82,14 @@ impl BatchOperationSqlType {
         }
     }
 
+    /// Whether this is an array type. Array columns can't go through
+    /// `(array_agg(col))[1]` in GROUP BY aggregation (array_agg over arrays
+    /// adds a dimension, so `[1]` yields NULL); they use MAX instead.
+    pub fn is_array(&self) -> bool {
+        matches!(self, BatchOperationSqlType::TextArray)
+            || matches!(self, BatchOperationSqlType::Custom(t) if t.contains("[]"))
+    }
+
     /// Returns the PostgreSQL type for binary COPY operations.
     pub fn to_pg_type(self) -> PgType {
         match self {
@@ -169,6 +177,10 @@ pub struct DynamicColumnDefinition {
     pub sql_type: BatchOperationSqlType,
     pub behavior: BatchOperationColumnBehavior,
     pub action: BatchOperationAction,
+    /// Starting value when an arithmetic upsert (Add/Subtract) creates the row,
+    /// so a first-touch `subtract` yields `default - value` instead of `+value`.
+    /// Must be a validated numeric literal — it is embedded directly in SQL.
+    pub insert_default: Option<String>,
 }
 
 impl DynamicColumnDefinition {
@@ -179,6 +191,11 @@ impl DynamicColumnDefinition {
         behavior: BatchOperationColumnBehavior,
         action: BatchOperationAction,
     ) -> Self {
-        Self { name, table_column: None, value, sql_type, behavior, action }
+        Self { name, table_column: None, value, sql_type, behavior, action, insert_default: None }
+    }
+
+    pub fn with_insert_default(mut self, insert_default: Option<String>) -> Self {
+        self.insert_default = insert_default;
+        self
     }
 }
--- a/core/src/database/clickhouse/generate.rs
+++ b/core/src/database/clickhouse/generate.rs
@@ -64,10 +64,11 @@ pub fn generate_tables_for_indexer_clickhouse(
             sql.push_str(format!("CREATE DATABASE IF NOT EXISTS {};", schema_name).as_str());
             info!("Creating database if not exists: {}", schema_name);
 
-            // Only create raw event tables for events in include_events (not for table-only events)
+            // Create raw event tables for every event whose raw events the
+            // runtime stores (include_events plus table-driving events)
             let raw_events: Vec<_> = event_names
                 .iter()
-                .filter(|e| contract.is_event_in_include_events(&e.name))
+                .filter(|e| contract.stores_raw_events(&e.name))
                 .cloned()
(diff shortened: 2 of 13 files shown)


ExoMonk:
@joshstevens19 

Reviewed + validated end-to-end on real Polygon data. 
Reproduced the bug with a synthetic add/subtract balances table over USDC.e transfers on a mid-history
window (87,780,000-87,780,150): the v0.42.0 binary diverges from per-holder net recomputed from raw transfer rows on 67 holders (of 83 whose first touch in the window is a debit); this branch matches the oracle exactly (0 mismatches), with raw event ingestion byte-identical across both binaries. Also regression-checked a set+min-only production config over a TokenRegistered-dense window (41.1M-41.2M, 29 reciprocal same-tx pairs): all table content byte-identical
across binaries, confirming set/min paths are untouched. Algebra holds for non-zero defaults, and the EXCLUDED rewrite inside pushed-down if: conditions keeps the raw-delta semantics. LGTM.

ExoMonk (approved):

PR #473, opened 2026-09-22
feat(hypersync): live indexing with hypersync (opt in)

DenhamPreen:
Adds `networks[].hypersync.head` (default false). 

Currently possible to use hypersync for speedy historical sync as implemented by @moose-code - #457 

This PR allows for opt in usage at the head, the value of using hypersync at the head is rooted in the cryptographic verification hypersync does to ensure data integrity of eth_getLogs highlighted in this [article](https://docs.envio.dev/blog/nodes-silently-miss-events). Shorter than the article -`eth_getLogs` response carries no proof of its own completeness.

The tradeoff is head latency, and I measured this. On Polygon I ran four head feeds side by side for three 3-minute windows (~120 blocks each) and recorded the first moment each feed reported each new block number:

- dRPC (paid, load-balanced endpoint), `eth_blockNumber` over HTTP polled every 100ms
- dRPC, `eth_subscribe("newHeads")` over websocket
- HyperSync, `GET /height` polled every 100ms
- HyperSync, `GET /height/sse` (what this PR uses)

Then 25 rounds of request round trips from the same machine: `eth_getBlockByNumber` vs a HyperSync single block query, and `eth_getLogs` over 100 blocks of the Polymarket CTF Exchange vs the same HyperSync log query.

What came out:

- HyperSync's SSE push trailed the 100ms dRPC poll by ~150ms at the median (one window had it 46ms ahead) and by 0.4–0.75s at p99. That gap is HyperSync ingesting and validating the block on top of the same upstream.
- Against dRPC's own push (websocket `newHeads`) it was on par: ahead on 92 of 119 blocks in one window, ~330ms behind at the median in another. dRPC's websocket delivery moved by ±700ms between sessions; the SSE feed was the steadier of the two.
- Polling HyperSync's `/height` was the worst of the four, 50–230ms behind its own SSE push at the median. That is why the PR uses the stream and not the existing 2s-cached poll.
- Request round trips were about half of dRPC's from my location when HyperSync was healthy (height 156 vs 285ms, logs ~195 vs ~400ms at p50), with one window where HyperSync queries had a slow tail (p90 over 1.2s).

So the trade is a few hundred milliseconds at the head for a data completeness. Happy to share the node script for benchmarking if useful.

The tip header still comes from RPC, so reorg detection, `reorg_safe_distance` and the bloom shortcut are unchanged.


DZakh:
Instead of `head` and `head_wait_ms`, how much harder would it be to add `for: backfill | realtime | backfill_and_realtime`? And in realtime mode, always run HyperSync and RPC in parallel without delay?

Although this's an important behavior decision that @joshstevens19 should decide, but I think it's a better default for users.

You can also check the logic to automatically switch between SSE and polling at HyperIndex https://github.com/enviodev/hyperindex/blob/main/packages/envio/src/sources/HeightFeed.res

